PR #6, opened 2017-02-20
- docs update to explain this configuration

srinatar:
- replaced WebSubnet with LBSubnet to minimize the Internet exposure


craigcarl-oracle:
@srinatar This example isn't specific to a load balanced workload, it's more generic. If you want a LBaaS specific example we can build one but it needs to be net new.

Issue #31, opened 2017-03-15
Provider should source credentials from the config file - 


craigcarl-oracle:
If ~/.oraclebmc/config exists and the configuration doesn't specify alternative credentials then the provider should source credentials from .oraclebmc/config.

Example config file is here - https://docs.us-phoenix-1.oraclecloud.com/Content/API/Concepts/sdkconfig.htm

Config files support the concept of profiles. We should support a variable (obmcs_profile??) that can optionally be set to the profile to be used. If a TF configuration doesn't specify a profile then use [DEFAULT].

craigcarl-oracle:
Here's a way we can do this outside of the provider, I think direct integration would be nicer.

https://gist.github.com/craigcarl-oracle/f00b14f86fc950fcdf14ea9a3e0136a9

josephholsten:
@jhorwit2  pointed me at https://gist.github.com/jhorwit2/7f18873988fb6496592e280c7b56f699

craigcarl-oracle:
We've decided this functionality should go in the SDK, not the provider. Once the re-factored SDK ships we'll re-evaluate.

PR #110, opened 2017-05-18
Issue #31: Provider should source credentials from the config file

derekwyatt:
- Added a function that will read the INI file and pull out a value for
  a specific key
- Created a function that will allow us to chain together multiple
  strategies for retreiving defaults
- Environment variables are preferred, falling back to the config file
  if needed

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -3,6 +3,7 @@
 package main
 
 import (
+	"os"
 	"errors"
 	"fmt"
 	"net/http"
@@ -10,6 +11,7 @@ import (
 	"github.com/MustWin/baremetal-sdk-go"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
+	"github.com/go-ini/ini"
 )
 
 var descriptions map[string]string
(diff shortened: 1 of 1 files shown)


codycushing:
Hello, I'm not seeing this actually work--removing TF_VAR_tenancy_ocid, TF_VAR_user_ocid, etc as exported values from the shell profile just causes terraform to prompt for those values in the command line. Can you describe the environment/plan configuration that allows verification of this functionality?

provider.go
```diff
@@ -38,25 +40,88 @@ func Provider(configfn schema.ConfigureFunc) terraform.ResourceProvider {
...
 	return map[string]*schema.Schema{
 		"tenancy_ocid": {
 			Type:        schema.TypeString,
 			Required:    true,
 			Description: descriptions["tenancy_ocid"],
-			DefaultFunc: schema.EnvDefaultFunc("OBMCS_TENANCY_OCID", nil),
+			DefaultFunc: firstOf(
+				"tenancy_ocid",
+				schema.EnvDefaultFunc("OBMCS_TENANCY_OCID", nil),
+				fileBasedConfig(os.Getenv("HOME") + "/.oraclebmc/config", "tenancy")),
```
codycushing: you would want to use `path.Join` for these and define `.oraclebmc/config` as a const.

provider.go
```diff
@@ -38,25 +40,88 @@ func Provider(configfn schema.ConfigureFunc) terraform.ResourceProvider {
...
+		return nil, fmt.Errorf("Unable to retrieve default for '%s' from anywhere", key)
+	}
+}
+
+// Look for default values in the BMC configuration file
+func fileBasedConfig(file string, key string) schema.SchemaDefaultFunc {
+	return func() (interface{}, error) {
+		cfg, err := ini.InsensitiveLoad(file)
+		if cfg != nil && err == nil {
+			sec, err := cfg.GetSection("")
```
codycushing: This looks like it will not work with a typical ini file that has named sections like 
[DEFAULT]
...
[USER1]
...

derekwyatt:
I think your typical and my typical are quote different (everyone I know
has only the default at this point) but you're certainly correct that
that's all it handles.

What would be the best way to allow the user to convey the different
section of the INI file to use? Or would you rather see an explicit
reference to "default"?

Barring a simple solution, this does work fine for me and should for anyone
else with only a single default section. Do you think it would be merge
able as is and defer the specification of the section to a future issue?

codycushing:
Hi @derekwyatt I never saw an answer as to how you are verifying functionality:
> I'm not seeing this actually work--removing TF_VAR_tenancy_ocid, TF_VAR_user_ocid, etc as exported values from the shell profile just causes terraform to prompt for those values in the command line. Can you describe the environment/plan configuration that allows verification of this functionality?

derekwyatt:
Sorry I missed that. I verified this by removing said variables. Is there
an established automated way of testing this?

codycushing:
Thanks for building this. Unfortunately due to this particular 3rd party library use we wont be able to merge. But, we have included plans to support ini based configs in the near future.

derekwyatt:
What 3rd partly library?  I added a dependency on `go-ini` but that's not new for this project.

gliptak:
Instead of ```INI``` file format, could we use "regular" ```tf``` format like other providers do? Thanks

https://www.terraform.io/docs/providers/aws/index.html#shared-credentials-file

derekwyatt:
The issue is that the Oracle BMCS config format is an INI file so the point
was to use that.

gliptak:
@derekwyatt I see. Reusing is good. Thank you for the clarification.

provider.go
josephholsten: would it be useful to have an env var to specify this, similar to `AWS_SHARED_CREDENTIALS_FILE`? Personally I don't care about it for regular users, just want to be override this path so I can add a test to this.

josephholsten:
@derekwyatt would you have a chance to rebase onto master? or would you prefer I just create a new PR updating this?

josephholsten:
well, I'm looking at how the AWS provider implements this, looks like they handle it in the SDK instead of the provider. I'm going to do more research, but it certainly seems like every SDK user would want this. But if it's going to be a while before the SDK is open source, this may the best way to solve #31 today.

derekwyatt:
Sorry... I was on vacation last week.  Do you still need me to rebase this?

codycushing:
This will be implemented at the SDK level with a custom library so it functions consistently with other OCI SDKs. Thanks

PR #124, opened 2017-05-24
Pass in a delay so that Terraform will wait until refreshing the reso…

jwolski2:
…urce

This is the first potential work-around for the compartment creation bug.

### DIFF
--- a/crud/helpers.go
+++ b/crud/helpers.go
@@ -27,6 +27,11 @@ var (
 	}
 )
 
+const (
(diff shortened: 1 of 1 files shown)


mikejihbe:
There's already a facility for this: `ExtraWaitPostCreateDelete`. Regardless, I think @craigcarl-oracle and friends at oracle are dealing with this in the API. This is a consistency problem on the backend, not something clients should have to worry about.

PR #138, opened 2017-06-02
Eginez/public buckets

eginez:
Adds support for access types to objects in buckets. Tested by running the plan under docs/examples/storage/object

### DIFF
--- a/docs/examples/storage/object/README.md
+++ b/docs/examples/storage/object/README.md
@@ -5,7 +5,7 @@
     #    \___/|_| \_/_/   \_\____|_____|_____|
 ***
 ## Creating buckets
-This example creates a single bucket inside a comparment
+This example creates a single bucket inside a compartment
 
 ### Using this example
 * Update env-vars with the required information. Most examples use the same set of environment variables so you only need to do this once.
--- a/docs/examples/storage/object/README.md
+++ b/docs/examples/storage/object/README.md
@@ -1,0 +1,34 @@
+    #     ___  ____     _    ____ _     _____
+    #    / _ \|  _ \   / \  / ___| |   | ____|
+    #   | | | | |_) | / _ \| |   | |   |  _|
+    #   | |_| |  _ < / ___ | |___| |___| |___
+    #    \___/|_| \_/_/   \_\____|_____|_____|
+***
+## Creating buckets
+This example creates a single bucket inside a comparment
+
+### Using this example
+* Update env-vars with the required information. Most examples use the same set of environment variables so you only need to do this once.
+* Source env-vars
+  * `$ . env-vars`
+* Update `variables.tf` with your instance options.
+
+### Files in the configuration
+
+#### `env-vars`
+Is used to export the environmental variables used in the configuration. These are usually authentication related, be sure to exclude this file from your version control system. It's typical to keep this file outside of the configuration.
+
+Before you plan, apply, or destroy the configuration source the file -  
+`$ . env-vars`
+
+#### `variables.tf`
+Defines the variables used in the configuration
+
+#### `object.tf`
+Defines the buckets we are going to create
+
+#### `outputs.tf`
+Defines the outputs of the configuration
+
+#### `provider.tf`
+Specifies and passes authentication details to the OBMCS TF provider
--- a/docs/examples/storage/object/README.md
+++ b/docs/examples/storage/object/README.md
@@ -27,8 +27,5 @@ Defines the variables used in the configuration
 #### `object.tf`
 Defines the buckets we are going to create
 
-#### `outputs.tf`
-Defines the outputs of the configuration
-
 #### `provider.tf`
 Specifies and passes authentication details to the OBMCS TF provider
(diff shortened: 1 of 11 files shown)


helpers_objectstorage.go
```diff
@@ -28,6 +30,11 @@ var bucketSchema = map[string]*schema.Schema{
 		Required: true,
 		Computed: false,
 	},
+	"access_type": {
+		Type:     schema.TypeString,
+		Computed: false,
+		Optional: true,
```
josephholsten: probably want to add `Default: baremetal.NoPublicAccess,`

helpers_objectstorage.go
```diff
@@ -28,6 +30,11 @@ var bucketSchema = map[string]*schema.Schema{
 		Required: true,
 		Computed: false,
 	},
+	"access_type": {
+		Type:     schema.TypeString,
+		Computed: false,
```
josephholsten: `Computed: false,` is the default. But it's included elsewhere, so at least it's consistent!

resource_obmcs_objectstorage_bucket.go
```diff
@@ -86,6 +87,12 @@ func (s *BucketResourceCrud) Create() (e error) {
 		metadata := resourceObjectStorageMapToMetadata(rawMetadata.(map[string]interface{}))
 		opts.Metadata = metadata
 	}
+	if accessType, ok := s.D.GetOk("accessType"); ok {
```
josephholsten: it's usually better to skip these bits, depend on a default in the schema and end up with something like:

```go
opts.AccessType = baremetal.BucketAccessType(s.D.Get("accessType"))
```

resource_obmcs_objectstorage_bucket.go
```diff
@@ -107,6 +114,12 @@ func (s *BucketResourceCrud) Update() (e error) {
 		opts.Metadata = metadata
 	}
 
+	if accessType, ok := s.D.GetOk("accessType"); ok {
```
josephholsten: meh, we really should extract this common data binding code from Create() & Update() someday.

resource_obmcs_objectstorage_object_test.go
```diff
@@ -41,9 +41,13 @@ func (s *ResourceObjectstorageObjectTestSuite) SetupTest() {
 	s.TimeCreated = baremetal.Time{Time: time.Now()}
 
 	s.Config = `
+		variable "accesstype" {
+			default = "ObjectRead"
+		}
 		resource "baremetal_objectstorage_bucket" "t" {
 			compartment_id = "${var.compartment_id}"
 			name = "bucketID"
+			access_type="${var.accesstype}"
```
josephholsten: is there any reason not to just inline:
```hcl
access_type = "ObjectRead"
```

I also like to run my configs through `terraform fmt`, but clearly that wasn't the case when you came to the code.

resource_obmcs_objectstorage_bucket.go
josephholsten: Woah! this absolutely needs to be using key `"access_type"` not `"accessType"`.

helpers_objectstorage.go
josephholsten: oh, while we're at it we probably want to add something like `ValidationFunc: validation.StringInSlice([]string{baremetal.NoPublicAccess, baremetal.ObjectRead}, true)`, see https://godoc.org/github.com/hashicorp/terraform/helper/validation#StringInSlice

resource_obmcs_objectstorage_bucket.go
eginez: nice catch, last minute refactoring! :/

helpers_objectstorage.go
eginez: this is great, thanks for the pointer

codycushing:
Can you update bucket.md with the new param

docs/examples/storage/object/README.md
```diff
@@ -0,0 +1,31 @@
+    #     ___  ____     _    ____ _     _____
+    #    / _ \|  _ \   / \  / ___| |   | ____|
+    #   | | | | |_) | / _ \| |   | |   |  _|
+    #   | |_| |  _ < / ___ | |___| |___| |___
+    #    \___/|_| \_/_/   \_\____|_____|_____|
+***
+## Creating buckets
+This example creates a single bucket inside a comparment
```
codycushing: comparment -> compartment

docs/examples/storage/object/variables.tf
```diff
@@ -0,0 +1,11 @@
+variable "tenancy_ocid" {}
+variable "user_ocid" {}
+variable "fingerprint" {}
+variable "private_key_path" {}
+variable "namespace_name" {}
+variable "compartment_ocid" {}
+
+variable "BucketName" {
+    default = "BucketOne"
+}
```
codycushing: Defining this variable only sort of complicates the example, guessing it was just a test

eginez:
Thanks for the input Cody, latest commit should have your comments addressed

eginez:
wohooo!

PR #152, opened 2017-06-16
Eginez/preauthreq

eginez:
Hey guys is the preauth req feature in the provider, test pass the only thing missing are docs, which I'll be updating soon.
This commit does not update the vendor folder. I can add those changes but I have a feeling the lb-fixes branch will add before I do and I'd like to avoid conflicts :D
Also this branch is based out of master, with the lb-fixes applied on top of it, so we would probably need the lb-fixes branch merged in before I merge this branch


eginez:
replaced with new PR

PR #154, opened 2017-06-16
console_history: fix length in acc-test

josephholsten:


data_source_obmcs_core_console_history_data_test.go
```diff
@@ -38,7 +38,7 @@ func (s *CoreConsoleHistoryDataDatasourceTestSuite) SetupTest() {
     }
     data "baremetal_core_console_history_data" "s" {
       console_history_id = "${baremetal_core_console_history.t.id}"
-      length = 1
+      length = 10240
```
codycushing: This number is a little magical, a comment for test maintainability would be welcome.

josephholsten: I agree. Probably should add a validator. That's just what the API error told me the min was.

PR #157, opened 2017-06-20
Adds load balancer examples

codycushing:


docs/examples/load_balancer/lb_lite/cert.key
```diff
@@ -0,0 +1,52 @@
+-----BEGIN PRIVATE KEY-----
```
rcohenma: is it safe to upload this unencrypted private key?

codycushing: This is a dummy cert created for examples.

PR #159, opened 2017-06-20
instance: fix round-trip serialization

### LINKED ISSUE #113: Terraform refresh is not getting remote state ?
## Description of the bug:
I'm creating full VCP stack and some test instances. If I remove VM instance via API/UI terraform doesn't catch this (other items like IPSec works).

terraform apply
```
baremetal_core_instance.TFInstance3: Refreshing state... (ID: ocid1.instance.oc1.phx.abyhqljsgqgxzl6o...3tu326kvuatk5bwbzsmhoeiodlazkmph2fc4gq)
No changes. Infrastructure is up-to-date.
```

this OCID is terminated instance.

josephholsten:
builds on #151, attempting to fix #113

### DIFF
--- a/data_source_obmcs_core_console_history_data.go
+++ b/data_source_obmcs_core_console_history_data.go
@@ -4,11 +4,12 @@ package main
 
 import (
 	"fmt"
+	"time"
+
 	"github.com/MustWin/baremetal-sdk-go"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/oracle/terraform-provider-baremetal/client"
 	"github.com/oracle/terraform-provider-baremetal/crud"
-	"time"
 )
 
 func ConsoleHistoryDataDatasource() *schema.Resource {
--- a/data_source_obmcs_core_instance_test.go
+++ b/data_source_obmcs_core_instance_test.go
@@ -1,0 +1,66 @@
+// Copyright (c) 2017, Oracle and/or its affiliates. All rights reserved.
+
+package main
+
+import (
+	"testing"
+
+	"github.com/MustWin/baremetal-sdk-go"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/terraform"
+
+	"github.com/oracle/terraform-provider-baremetal/crud"
+)
+
+func TestIsStatefulResource(t *testing.T) {
+	var sr crud.StatefulResource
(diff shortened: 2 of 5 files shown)


josephholsten:
@codycushing what config are you using? I'm not seeing that with the [config in the acc test](https://github.com/oracle/terraform-provider-baremetal/pull/159/files#diff-1b915eeef9abc59aade8b80c803516daR57):

```hcl
resource "baremetal_core_instance" "t" {
  availability_domain = "${data.baremetal_identity_availability_domains.ADs.availability_domains.0.name}"
  compartment_id      = "${var.compartment_id}"
  image               = "` + FastestImageID + `"
  shape               = "` + SmallestShapeName + `"
  subnet_id           = "${baremetal_core_subnet.s.id}"
}
```

codycushing:
It's with a plan using the nested vnic object:
```
resource "baremetal_core_instance" "instance1" {
  availability_domain = "${lookup(data.baremetal_identity_availability_domains.ADs.availability_domains[0],"name")}" 
  compartment_id = "${var.compartment_ocid}"
  display_name = "be-instance1"
  image = "${lookup(data.baremetal_core_images.image-list.images[0], "id")}"
  shape = "VM.Standard1.1"

  create_vnic_details {
    subnet_id = "${var.subnet}"
    display_name = "vnic1"
    assign_public_ip = true
    hostname_label = "be-instance1"
  },
}
```

josephholsten:
@codycushing I see what's happening now!

essentially, after the instance is created, the impl is wanting to set `resource.baremetal_core_instance.instance1.subnet_id` from the response. I was hoping https://github.com/oracle/terraform-provider-baremetal/pull/159/files#diff-930932b2c63a550ddecdf09c110553bdR242 would keep that from breaking things, but I'd actually need to hack the state (and violate layering) in order to trick terraform into thinking that a null `resource.baremetal_core_instance.instance1.subnet_id` doesn't need to change.

Talking with @grubernaut, it looks like it's not really feasible to support the old `resource.baremetal_core_instance.instance1.subnet_id` attribute and the new `resource.baremetal_core_instance.instance1.create_vnic_details.0.subnet_id` in the same impl, so we need a backwards incompatible change.

While we're at it, I'd really like to rename the `create_vnic_details` attribute to `vnic`, which makes more sense when reading attributes, and will be more clear when (if?) instances support more than one vnic.

josephholsten:
I'm going to create a variant of this PR that is backward incompatible, but has state migration so no data (or resources) are broken

PR #165, opened 2017-06-28
Load Balancer supports session persistence and local-vcn

codycushing:
* adds example showing private lb and session persistence
* fix bad doc for backendset


client/bare_metal_client.go
```diff
@@ -12,7 +12,7 @@ type BareMetalClient interface {
 	CaptureConsoleHistory(instanceID string, opts *baremetal.RetryTokenOptions) (icHistory *baremetal.ConsoleHistoryMetadata, e error)
 
 	CreateBackend(loadBalancerID string, backendSetName string, ipAddr string, port int, opts *baremetal.CreateLoadBalancerBackendOptions) (workRequestID string, e error)
-	CreateBackendSet(loadBalancerID string, name string, policy string, backends []baremetal.Backend, healthChecker *baremetal.HealthChecker, sslConfig *baremetal.SSLConfiguration, opts *baremetal.LoadBalancerOptions) (workRequestID string, e error)
+	CreateBackendSet(loadBalancerID string, name string, policy string, backends []baremetal.Backend, healthChecker *baremetal.HealthChecker, sslConfig *baremetal.SSLConfiguration, sessionPersistenceConfig *baremetal.SessionPersistenceConfiguration, opts *baremetal.LoadBalancerOptions) (workRequestID string, e error)
```
rcohenma: Need to update the function signatures in the mock client too

data_source_obmcs_core_console_history_data.go
```diff
@@ -36,7 +35,7 @@ func ConsoleHistoryDataDatasource() *schema.Resource {
 					}
 
 					if v < 10240 {
-						es = append(es, fmt.Errorf("expected %s to be in the range (%d - %d), got %d", k, min, max, v))
+						es = append(es, fmt.Errorf("expected %s to be less than %d, got %d", k, 10240, v))
```
rcohenma: 10240 could be put in a variable so that it doesn't have to be changed in multiple places

Issue #175, opened 2017-07-11
No Request URL for PRA


rgmccaw:
## Description of the bug:

For the `baremetal_objectstorage_preauthrequest` resource there is no Pre-Authorized Request URL in the attributes available for output.

The url is needed in order to make use of the PRA.

## A self-contained terraform file that reproduces the issue:

```
#     ___  ____     _    ____ _     _____
#    / _ \|  _ \   / \  / ___| |   | ____|
#   | | | | |_) | / _ \| |   | |   |  _|
#   | |_| |  _ < / ___ | |___| |___| |___
#    \___/|_| \_/_/   \_\____|_____|_____|

## PAR(preauthenticated requests) management.
## This example creates two preauthenticated requests, one on a bucket and the second on an object
## Export the environmental variables used for configuration.
## Provide the script with an object and bucket present in your namespace

variable "tenancy_ocid" {}
variable "user_ocid" {}
variable "fingerprint" {}
variable "private_key_path" {}
variable "compartment_ocid" {}
variable "namespace_name" {}

variable "bucket_name" {}
variable "object_name" {}

provider "baremetal" {
  region = "us-phoenix-1"
  alias = "phx"
  tenancy_ocid = "${var.tenancy_ocid}"
  user_ocid = "${var.user_ocid}"
  fingerprint = "${var.fingerprint}"
  private_key_path = "${var.private_key_path}"
}

resource "baremetal_objectstorage_preauthrequest" "parOnObject" {
  namespace = "${var.namespace_name}"
  bucket = "${var.bucket_name}"
  object = "${var.object_name}"
  name = "parOnObject"
  access_type = "ObjectRead" //Other configurations accepted are ObjectWrite, ObjectReadWrite
  time_expires = "2019-11-10T23:00:00Z"
}

output "pra_request_url" {
  value = "${baremetal_objectstorage_preauthrequest.parOnObject.access_uri}" # access_uri or similar does not exist
}
```

eginez:
Hey there thanks for the bug, I am little confuse.
Is it that the url is `acces_uri` is empty? or is it that the access_uri complete

eginez:
Never mind I see what you mean that schema is missing the "access_uri" field

codycushing:
@eginez Im not seeing the `access_uri` defined in `preauthenticatedRequestSchema`, it would be a computed value.

rgmccaw:
Issue is fixed in release v1.0.15

[...]

PR #177, opened 2017-07-12
adding missing access_uri to schema #175

eginez:


eginez:
test with ``` terraform apply```

helpers_objectstorage.go
```diff
@@ -110,6 +110,11 @@ var preauthenticatedRequestSchema = map[string]*schema.Schema{
 			string(baremetal.PARObjectReadWrite),
 			string(baremetal.ObjectRead)}, true),
 	},
+	"access_uri": {
```
rcohenma: Access_type is required for create according to the spec 
https://bitbucket.aka.lgl.grungy.us/projects/CASPER/repos/casper-api-spec/browse/src/main/resources/casper-api.yaml

docs/examples/storage/preauthrequests/preauthreq.tf
```diff
@@ -20,8 +20,6 @@ variable "bucket_name" {}
 variable "object_name" {}
 
 provider "baremetal" {
-  region = "us-phoenix-1"
-  alias = "phx"
```
codycushing: If you define a provider with an alias, resources must specify that alias, otherwise it appears terraform is attempting to instantiate its own bmcs provider, which would be why it then prompts for all the required params but fails because both key path and key material appear to be optional but in reality one or the other is required.

eginez: got it!

helpers_objectstorage.go
eginez: addressed see above

PR #178, opened 2017-07-12
added extended_metadata field to instance

rcohenma:


docs/datasources/core/instances.md
```diff
@@ -36,6 +36,7 @@ The following attributes are exported:
 * `image_id` - The image used to boot the instance. You can enumerate all available images by calling ListImages.
 * `state` - The current state of the instance: [PROVISIONING, RUNNING, STARTING, STOPPING, STOPPED, CREATING_IMAGE, TERMINATING, TERMINATED]
 * `metadata` - Custom metadata that you provide.
+* `extended_metadata` - Custom nested metadata that you provide.
```
codycushing: Recommend adding a note on how json vs invalid json is handled and represented in the system.

data_source_obmcs_core_instance.go
```diff
@@ -161,3 +168,18 @@ func (s *InstanceDatasourceCrud) SetData() {
 	}
 	return
 }
+
```
codycushing: Please test if this is still necessary and if so add a brief note explaining the problem and how this solves it.

docs/examples/compute/extended_metadata/extended_metadata.tf
```diff
@@ -0,0 +1,96 @@
...
+        connection {
+            agent = false
+            timeout = "30m"
+            host = "${data.baremetal_core_vnic.InstanceVnic.public_ip_address}"
+            user = "opc"
+            private_key = "${var.ssh_private_key}"
+        }
+        inline = [
+            "export STR=$(curl http://169.254.169.254/opc/v1/instance/metadata/nested_object/object/some_string)",
+            "if [[ \"$STR\" != \"stringC\" ]]; then exit 1; fi"
```
codycushing: Cool! Linking to the docs on accessing extended metadata might be good here

docs/examples/compute/extended_metadata/extended_metadata.tf
```diff
@@ -0,0 +1,96 @@
...
+    default = "VM.Standard1.2"
+}
+
+variable "InstanceOS" {
+    default = "Oracle Linux"
+}
+
+variable "InstanceOSVersion" {
+    default = "7.3"
+}
```
codycushing: I've been simplifying new examples to put the above image related vars inside the baremetal_core_images datasource

data_source_obmcs_core_instance.go
rcohenma: it is necessary because in the schema above we had to specify the extended_metadata as a map[string]string
This is because afaik there is no way to specify a nested map of arbitrary depth in the schema

PR #183, opened 2017-07-19
adding option to disable automatic retries

rcohenma:


eginez:
looks good, it would be great to add a brief example on how to use this

eginez:
great thanks approved

brendanhowes:
@rcohenma @codycushing is the disable retries still needed (as of 2.0.7)? its still in docs/examples/compute/instance/provider.tf

PR #181, opened 2017-07-17
Add TravisCI build file

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

https://github.com/oracle/terraform-provider-baremetal/issues/166
https://github.com/oracle/terraform-provider-baremetal/issues/171

https://travis-ci.org/gliptak/terraform-provider-baremetal/jobs/254643997

```
client/mocks/bare_metal_client.go:3709: cannot use (*BareMetalClient)(nil) (type *BareMetalClient) as type client.BareMetalClient in assignment:

	*BareMetalClient does not implement client.BareMetalClient (wrong type for CreateLoadBalancer method)

		have CreateLoadBalancer(*baremetal.BackendSet, *baremetal.Certificate, string, *baremetal.Listener, string, []string, *baremetal.CreateOptions) (string, error)

		want CreateLoadBalancer(*baremetal.BackendSet, *baremetal.Certificate, string, *baremetal.Listener, string, []string, *baremetal.CreateLoadBalancerOptions) (string, error)
```

PR #184, opened 2017-07-20
Add Wercker CI/CD

tjfontaine:
This adds Wercker.com CI/CD: https://app.wercker.com/tjfontaine/terraform-provider-baremetal/runs

Currently `go vet` is disabled because of unreachable code: https://github.com/oracle/terraform-provider-baremetal/blob/v1.0.16/resource_obmcs_identity_compartment.go#L59-L60

`go test` is also disabled because they require credentials (which can be enabled in Wercker) but I didn't have access to them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ terraform-encrypted-des.sh
 terraform-encrypted.sh
 terraform-plain.sh
 .DS_Store
+
+.wercker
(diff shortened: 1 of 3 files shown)


gliptak:
https://github.com/oracle/terraform-provider-baremetal/pull/181

brunoborges:
Hi @gliptak, moving forward we should prioritize the use of Wercker whenever it is possible. If for some reason it doesn't work for CI/CD of a specific project, then we can use Travis.

brunoborges:
Cool. PR merged, build triggered, and badge added to readme on commit 197b926.

See more: https://app.wercker.com/Oracle/terraform-provider-baremetal/

PR #199, opened 2017-07-28
Enable go vet in Wercker

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

### DIFF
--- a/resource_obmcs_identity_compartment.go
+++ b/resource_obmcs_identity_compartment.go
@@ -3,7 +3,6 @@
 package main
 
 import (
-	"fmt"
 	"strings"
 
 	"github.com/MustWin/baremetal-sdk-go"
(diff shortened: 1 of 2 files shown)


resource_obmcs_identity_compartment.go
```diff
@@ -56,7 +56,7 @@ func updateCompartment(d *schema.ResourceData, m interface{}) (e error) {
 func deleteCompartment(d *schema.ResourceData, m interface{}) (e error) {
 	sync := &CompartmentResourceCrud{}
 	sync.D = d
-	return crud.DeleteResource(d, sync)
+	//return crud.DeleteResource(d, sync)
```
tjfontaine: it seemed to me the opposite was the desired case, the `fmt.Errorf` was the old version, and the `DeleteResource` was the intended current state. At least if I look at `updateCompartment` as an example.

gliptak: modified

PR #201, opened 2017-07-29
goimports

josephholsten:


codycushing:
Tested and works, did require cleaning and updating my x/tools/* source. To reduce unnecessary merge conflicts, we should do this change with Ramys rename task, I'll make sure it gets included there.

Makefile
```diff
@@ -10,7 +10,7 @@ clean:
 	rm -rf bin/*
 
 fmt:
-	gofmt -w $(GOFMT_FILES)
+	goimports -w -local github.com/oracle/terraform-provider-baremetal $(GOFMT_FILES)
```
codycushing: @rcohenma can you include this change in your current rename branch then run `make fmt`.

josephholsten:
I was just going to ask if anyone had any significant branches in progress ;-). I'll rip out the actual fmt change so we can apply it when things are less in flux.

josephholsten:
Updated. Should be boring to merge, and we can fmt when it's convenient.

PR #205, opened 2017-08-02
enabling exadata and updating DBSystem docs

rcohenma:


resource_obmcs_database_db_system.go
```diff
@@ -349,7 +349,6 @@ func (s *DBSystemResourceCrud) SetData() {
 	s.D.Set("cpu_core_count", s.Res.CPUCoreCount)
 	s.D.Set("database_edition", s.Res.DatabaseEdition)
 	s.D.Set("db_home", s.Res.DBHome)
-	s.D.Set("hostname", s.Res.Hostname)
```
codycushing: Merely removing this will likely lead someone else later to notice it's absent and add it again--instead we should comment out and add a note explaining the workaround for the atypical service behavior.

PR #211, opened 2017-08-04
vendor: rm unused

josephholsten:
Fixes #211


codycushing:
We'll do this all at once with some other large scale dependency updates

PR #214, opened 2017-08-07
fix for dbsystems attributes

rcohenma:
Cluster_name is auto generated if the user does not provide one. Without the computed flag set to true, the DBSystem would be destroyed and relaunched in the case where the user does not provide a cluster_name.
DBSystem does not have an update function according to their spec. Without the ForceNew on the DBHome and the Database parameters, the DBSystem would not be destroyed and relaunched in the case where the user changes one of the values and runs apply again.


codycushing:
Please include a basic commit description and test plan for verification

PR #215, opened 2017-08-09
test: rm mocks

josephholsten:
mocked tests were broken and mostly removed long ago, this removes the many remnants.

(also includes #201 so I don't have to rebase everything if it's merged)


josephholsten:
I've spit the commits into extremely granular pieces so we can be sure each one is correct, it looks like a mess when they're all together. That being said, it may make sense to squash them at merge.

josephholsten:
@codycushing @rcohenma the only functional changes in this are supposed to be in https://github.com/oracle/terraform-provider-baremetal/pull/215/commits/5ab2ab39c5bd1653e1fca0a4605dd518795bc26b, just lb test changes.

josephholsten:
woah woah woah. Getting a panic from this, bisecting it now.

josephholsten:
well shoot. https://github.com/oracle/terraform-provider-baremetal/pull/215/commits/1fdf90866faf97fe82986c9238e7cca4b89481a4 exposes brokenness that was just hiding. Guess I have to find out why these are broken.

josephholsten:
all right! turns out there were some null references in the original code, hidden by an `IsAccTest` check. They are now fixed by https://github.com/oracle/terraform-provider-baremetal/pull/215/commits/d5150e52dbe466144b9e8923363b039cb5464f3c

rcohenma:
Why are we removing the unit tests instead of fixing them?

josephholsten:
TL;DR: they don't work now, they're painful for our the use-cases they could help with, other options hurt less.

@rcohenma That's an excellent question, and the fact that this looks like I'm "removing unit tests" is part of the problem. I'm actually going to avoid the words "unit test" and "acceptance test" in my answer because I think they lead to confusion between the terms used the testing libs being used, and what those terms mean in the general sense.

Terraform provider implementations are required to have live, end-to-end tests in order to be accepted as an official provider. Most of the tests we currently have work properly as end-to-end tests when connected to a live backend. But currently there is a bunch of added complexity around these tests to theoretically support other modes of testing, but do not work at this moment.

One of the theoretical goals is to support offline testing. As seen in the SDK, that can make for a huge benefit in development. I certainly don't run the entire test suite live before proposing a PR, much less run it continually while developing. Unfortunately, abstracting the API at the SDK level in the tests requires a ton of maintenance continually, which none of us has been diligent about doing. I'd actually like to re-add offline testing, but instead of manually mocking the SDK calls, I think we'd have a more pleasant time by using recorded HTTP requests & responses as done in the SDK tests.

Another theoretical goal is to test that the correct interface calls are being made. As before, we haven't been diligent about maintaining these mock expectations, so we don't currently have that in practice. But I would also argue that using mock expectations in the tests designed to be end-to-end would make them much more complicated. Traditionally, mockist testing strives for one-test-one-check, and considers significant-test-setup a code smell. If we want this kind of testing, I'd really prefer to have tests of single functions, eg: [InstanceResourceCrud.SetData()](https://github.com/oracle/terraform-provider-baremetal/blob/master/resource_obmcs_core_instance.go#L372). Most other providers don't bother with this level of testing, but they don't have an abstraction layer like our `crud` package. 

A traditional goal of mockist testing is to be able to simulate errors from external dependencies. This is probably the most significant gap in our current testing, and the one I'm least able to suggest a solution to. It's certainly possible to stub some error responses through http mocking, as I'm suggesting for offline testing, but I don't know if that gets us where we'd like.

Does that sound like your experience? Have I gone crazy?

josephholsten:
oh, some of the community expectations around testing are mentioned in https://github.com/hashicorp/terraform/blob/master/docs/maintainer-etiquette.md

codycushing:
Maintaining the 3000 line boilerplate mock file for acceptance testing isn't practicable. It makes sense for unit testing where you are testing behavior at a function level, less so for real world end to end tests. For that we should mock at a higher level of abstraction with recorded data.

PR #220, opened 2017-08-10
test: replace resource.UnitTest with resource.Test

josephholsten:
test: replace resource.UnitTest with resource.Test

resource.UnitTest doc specifies:
>  This should only be used for resource that don't have any external dependencies.

See: https://godoc.org/github.com/hashicorp/terraform/helper/resource#UnitTest

Similarly, resource.TestCase doc specifies:
>  IsUnitTest allows a test to run regardless of the TF_ACC
> environment variable. This should be used with care - only for
>  fast tests on local resources (e.g. remote state with a local
> backend) but can be used to increase confidence in correct
>  operation of Terraform without waiting for a full acctest run.

See: https://godoc.org/github.com/hashicorp/terraform/helper/resource#TestCase

### DIFF
--- a/data_source_obmcs_core_console_history_data_test.go
+++ b/data_source_obmcs_core_console_history_data_test.go
@@ -49,7 +49,7 @@ func (s *CoreConsoleHistoryDataDatasourceTestSuite) TestResourceShowConsoleHisto
 	data := make([]byte, 100)
 	rand.Read(data)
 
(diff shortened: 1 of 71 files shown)


josephholsten:
It's possible that some of these can be run as UnitTests, but my review of them didn't indicate so.

codycushing:
Thanks, we'll convert as we touch each service groups tests

PR #221, opened 2017-08-11
Fix Console History tests

codycushing:
* refine instance fixtures to use fastest VM

TESTS:
* run `TF_ACC=1 TF_ORACLE_ENV=test ./terraform-plain.sh -timeout 120m -run TestResourceCoreConsoleHistoryTestSuite`
* run `TF_ACC=1 TF_ORACLE_ENV=test ./terraform-plain.sh -timeout 120m -run TestDatasourceCoreConsoleHistoryTestSuite`

### DIFF
--- a/provider_test.go
+++ b/provider_test.go
@@ -58,52 +58,14 @@ resource "baremetal_core_virtual_network" "t" {
 	display_name = "network_name"
 }
 
-resource "baremetal_core_internet_gateway" "CompleteIG" {
-    compartment_id = "${var.compartment_id}"
-    display_name = "CompleteIG"
-    vcn_id = "${baremetal_core_virtual_network.t.id}"
-}
-
-resource "baremetal_core_route_table" "RouteForComplete" {
-    compartment_id = "${var.compartment_id}"
-    vcn_id = "${baremetal_core_virtual_network.t.id}"
-    display_name = "RouteTableForComplete"
-    route_rules {
-        cidr_block = "0.0.0.0/0"
-        network_entity_id = "${baremetal_core_internet_gateway.CompleteIG.id}"
-    }
-}
-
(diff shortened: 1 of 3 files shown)


resource_obmcs_core_console_history.go
```diff
@@ -110,6 +110,11 @@ func (s *ConsoleHistoryResourceCrud) Get() (e error) {
 	return
 }
 
+func (s *ConsoleHistoryResourceCrud) Delete() (e error) {
+	e = nil
```
briangustafson: I'm probably missing some context. It looks like this would silently succeed without doing the delete?

codycushing: A few lines up in `deleteConsoleHistory` the `return fmt.Errorf("console history resource: console history %v cannot be deleted", d.Id())` line was causing the test to fail even after the test assertion succeeded. It also indicated the resource could not be deleted, this is the pattern for that. Looking at the api docs, there is a DeleteConsoleHistory api--probably added after this was developed a year ago, but I'm more inclined to remove this feature and tests than invest more in it--there is no use case for console history in terraform, it was originally added by mistake.

briangustafson: Makes sense - I agree that there isn't much of a use case here. If it's possible to log a warning here to say that delete isn't supported, that would be good.

provider_test.go
```diff
@@ -58,52 +58,14 @@ resource "baremetal_core_virtual_network" "t" {
 	display_name = "network_name"
 }
 
-resource "baremetal_core_internet_gateway" "CompleteIG" {
```
briangustafson: Why are these being removed?

codycushing: Not used in the test at all--added setup time for fixtures.

PR #237, opened 2017-09-12
adding support for Secondary Private IPs

rcohenma:


docs/examples/compute/private_ip/private_ip.tf
```diff
@@ -0,0 +1,77 @@
...
+    availability_domain = "${lookup(data.oci_identity_availability_domains.ADs.availability_domains[var.AD - 1],"name")}" 
+    compartment_id = "${var.compartment_ocid}"
+    display_name = "TFInstance"
+    hostname_label = "instance"
+    image = "${lookup(data.oci_core_images.OLImageOCID.images[0], "id")}"
+    shape = "VM.Standard1.2"
+    subnet_id = "${var.subnet_ocid}"
+}
+
+# Gets a list of vNIC attachments on the instance
```
briangustafson: VNIC

docs/examples/compute/private_ip/private_ip.tf
```diff
@@ -0,0 +1,77 @@
...
+# Create PrivateIP
+resource "oci_core_private_ip" "TFPrivateIP" {
+	vnic_id = "${lookup(data.oci_core_vnic_attachments.InstanceVnics.vnic_attachments[0],"vnic_id")}"
+	display_name = "someDisplayName"
+	hostname_label = "somehostnamelabel"
+}
+
+# List Private IPs
+data "oci_core_private_ips" "privateIpDatasource" {
+	depends_on = ["oci_core_private_ip.TFPrivateIP"]
```
briangustafson: Is it necessarily to have the depends_on when you use that same resource in the next line?

rcohenma: I found that it was in the case of the datasource

PR #248, opened 2017-09-15
Fix prohibit_public_ip_on_vnic

briangustafson:
Imports subnet's prohibit_public_ip_on_vnic property and forces
new when changed. In addition to the test update, also validated
that if a statefile and plan does not have this property set
(which would be the case for plans run before this property
existed), then upgrading and running plan will correctly compare the
value with the default and make no changes.

### DIFF
--- a/resource_obmcs_core_subnet_test.go
+++ b/resource_obmcs_core_subnet_test.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"testing"
 
+	"fmt"
 	"github.com/MustWin/baremetal-sdk-go"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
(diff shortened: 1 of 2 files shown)


resource_obmcs_core_subnet_test.go
```diff
@@ -17,15 +18,26 @@ func TestAccResourceCoreSubnetCreate_basic(t *testing.T) {
...
 
+	commonSubnetParams := `
+					availability_domain = "${data.oci_identity_availability_domains.ADs.availability_domains.0.name}"
+					compartment_id = "${var.compartment_id}"
+					vcn_id = "${oci_core_virtual_network.t.id}"
+					security_list_ids = ["${oci_core_virtual_network.t.default_security_list_id}"]
+					route_table_id = "${oci_core_virtual_network.t.default_route_table_id}"
+					dhcp_options_id = "${oci_core_virtual_network.t.default_dhcp_options_id}"
+					cidr_block = "10.0.2.0/24"`
+
```
codycushing: Lets discuss our approach to templatizing reusable config blocks, stuff in provider_test.go is not maintainable and we can probably come up with something like this but more systematic, modular and granular across the board.

PR #257, opened 2017-09-20
Removing resource from state if it has been terminated so that it is recreated on next apply

rcohenma:
…recreated on next apply

### DIFF
--- a/crud/helpers.go
+++ b/crud/helpers.go
@@ -235,8 +235,7 @@ func ReadResource(sync ResourceReader) (e error) {
 
 	sync.SetData()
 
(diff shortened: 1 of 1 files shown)


briangustafson:
Is there a specific bug that this is fixing? Can we write a test case for it?

rcohenma:
This is to fix issue #113 
not sure if it is possible to write a test like this. The case is an instance is created with a terraform plan, the instance then dies or is terminated using the UI.
This fix will make it so that the next apply would recreate the instance, rather than just refreshing the state to TERMINATED in the statefile.

PR #260, opened 2017-09-21
Adds VNIC skip_source_dest_check property

briangustafson:
Also sets changes to instance's create_vnic_details to ForceNew,
and adds a test for this.

Tests run:
make test run=TestDataSourceCoreVnicTestSuite debug=1
make test run=ResourceCoreVnicAttachmentTestSuite debug=1
make test run=ResourceCoreInstanceTestSuite debug=1

### DIFF
--- a/helpers_core.go
+++ b/helpers_core.go
@@ -32,5 +32,11 @@ func SetCreateVnicOptions(rawCreateVnicDetails interface{}) (vnicOpts *baremetal
 		*vnicOpts.AssignPublicIp = assignPublicIp.(string) == "1"
 	}
 
+	skipSourceDestCheck := vnic["skip_source_dest_check"]
+	if skipSourceDestCheck != nil {
+		vnicOpts.SkipSourceDestCheck = new(bool)
+		*vnicOpts.SkipSourceDestCheck = skipSourceDestCheck.(string) == "1"
+	}
+
 	return
 }
--- a/resource_obmcs_core_instance.go
+++ b/resource_obmcs_core_instance.go
@@ -32,27 +32,38 @@ func InstanceResource() *schema.Resource {
 			"create_vnic_details": {
 				Type:     schema.TypeMap,
 				Optional: true,
+				ForceNew: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"assign_public_ip": {
 							Type:     schema.TypeBool,
 							Optional: true,
+							ForceNew: true,
 						},
 						"display_name": {
 							Type:     schema.TypeString,
 							Optional: true,
+							ForceNew: true,
 						},
 						"hostname_label": {
 							Type:     schema.TypeString,
 							Optional: true,
+							ForceNew: true,
 						},
 						"private_ip": {
 							Type:     schema.TypeString,
 							Optional: true,
+							ForceNew: true,
+						},
+						"skip_source_dest_check": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							ForceNew: true,
 						},
 						"subnet_id": {
 							Type:     schema.TypeString,
 							Required: true,
+							ForceNew: true,
 						},
 					},
 				},
--- a/resource_obmcs_core_instance_test.go
+++ b/resource_obmcs_core_instance_test.go
@@ -9,9 +9,9 @@ import (
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
 
+	"fmt"
 	"github.com/oracle/terraform-provider-oci/crud"
 	"github.com/stretchr/testify/suite"
-	"fmt"
 )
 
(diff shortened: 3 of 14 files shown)


helpers_core.go
```diff
@@ -32,5 +32,11 @@ func SetCreateVnicOptions(rawCreateVnicDetails interface{}) (vnicOpts *baremetal
 		*vnicOpts.AssignPublicIp = assignPublicIp.(string) == "1"
 	}
 
+	skipSourceDestCheck := vnic["skip_source_dest_check"]
+	if skipSourceDestCheck != nil {
+		vnicOpts.SkipSourceDestCheck = new(bool)
+		*vnicOpts.SkipSourceDestCheck = skipSourceDestCheck.(string) == "1"
```
codycushing: Does this always come through as "1" or is it ever possibly "true"?

resource_obmcs_core_instance_test.go
```diff
@@ -109,6 +117,91 @@ func (s *ResourceCoreInstanceTestSuite) TestAccResourceCoreInstance_basic() {
...
+					func(ts *terraform.State) (err error) {
+						newId, err := fromInstanceState(ts, s.ResourceName, "id")
+						if newId != instanceId {
+							return fmt.Errorf("Expected same instance ocid, got the different.")
+						}
+						return err
+					},
+				),
+			},
+			// verify force refresh by creating an instance with create_vnic_details.
```
codycushing: Do you mean "force refresh" or "force new" or "recreated"?

resource_obmcs_core_instance.go
```diff
@@ -32,27 +32,38 @@ func InstanceResource() *schema.Resource {
...
 						},
 						"private_ip": {
 							Type:     schema.TypeString,
 							Optional: true,
+							ForceNew: true,
+						},
+						"skip_source_dest_check": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							ForceNew: true,
```
codycushing: Regarding display_name, hostname_label and skip_source_dest_check having _ForceNew: true_, as these fields can be updated through the vnic api, is there a reasonable expectation from a user that they should be able to update these values here without their instance being recreated? Can they be updated separately while attached to an instance? If so, does that have an ill-effect here when running plan/apply on an existing plan after?

resource_obmcs_core_instance_test.go
```diff
@@ -109,6 +117,91 @@ func (s *ResourceCoreInstanceTestSuite) TestAccResourceCoreInstance_basic() {
...
+					compartment_id = "${var.compartment_id}"
+					image = "${data.oci_core_images.t.images.0.id}"
+					shape = "VM.Standard1.1"
+					display_name = "-tf-instance"
+					metadata {
+						ssh_authorized_keys = "${var.ssh_public_key}"
+					}
+					create_vnic_details {
+							subnet_id = "${oci_core_subnet.t.id}"
+							display_name = "-tf-vnic"
```
codycushing: Extra tabs here

resource_obmcs_core_instance_test.go
```diff
@@ -109,6 +117,91 @@ func (s *ResourceCoreInstanceTestSuite) TestAccResourceCoreInstance_basic() {
...
+							assign_public_ip = true
+							private_ip = "10.0.1.20"
+							skip_source_dest_check = true
```
codycushing: extra tabs, be sure to run "go fmt" on your changed files prior to commit

helpers_core.go
briangustafson: Good question... this is, as mentioned, a workaround for https://github.com/hashicorp/terraform/issues/13512. However, we're not using this approach everywhere where we have bools. Let me do a bit of testing around this and get back to you.

resource_obmcs_core_instance_test.go
briangustafson: force new, updated

resource_obmcs_core_instance.go
briangustafson: Yes, yes, and yes. Basically, there's more work to be done here, but this change at least accurately reflects the current behavior. Currently, any change to these properties on an existing instance will be ignored. Forcing New on these will allow property updates to be applied. The remaining work should be:

1) Update the VNIC for the 3 properties that you mention.
2) Import/Refresh or create_vnic_details.

I'll open items to track these. They shouldn't be difficult, but are a bit strange in that VNIC management will need to happen within the instance resource.

The change in this PR is, I believe, an improvement, just not all the way to where we want to be.

resource_obmcs_core_instance_test.go
briangustafson: fixed

resource_obmcs_core_instance_test.go
briangustafson: fixed - go fmt doesn't actually catch this, I assume since it's inside of a string.

PR #263, opened 2017-09-22
Fix boolean parsing in create_vnic_details

briangustafson:
For most bool values in Terraform, you can use true and "true" interchangeably.
The bool values nested in create_vnic_details are treated differently, apprently
due to https://github.com/hashicorp/terraform/issues/13512: true appears as "1",
while "true" appears as "true". ParseBool handles both.

Verified that the changes to ResourceCoreVnicAttachmentTestSuite fail before applying the other changes and succeed after.

### DIFF
--- a/resource_obmcs_core_instance.go
+++ b/resource_obmcs_core_instance.go
@@ -260,16 +260,19 @@ func (s *InstanceResourceCrud) Create() (e error) {
 	}
 
 	if rawVnic, ok := s.D.GetOk("create_vnic_details"); ok {
-		opts.CreateVnicOptions = SetCreateVnicOptions(rawVnic)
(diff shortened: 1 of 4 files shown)


resource_obmcs_core_instance.go
```diff
@@ -260,16 +260,19 @@ func (s *InstanceResourceCrud) Create() (e error) {
 	}
 
 	if rawVnic, ok := s.D.GetOk("create_vnic_details"); ok {
-		opts.CreateVnicOptions = SetCreateVnicOptions(rawVnic)
+		opts.CreateVnicOptions, e = SetCreateVnicOptions(rawVnic)
+	}
+
+	if e == nil {
```
rcohenma: may look better to have an if  e != nil {return} right after the  SetCreateVnicOptions
but up to you

PR #264, opened 2017-09-23
Update references to Oracle owned sdk repo

codycushing:
* this is still a private repo but there is no change to the contents of the respective vendor codefiles

### Test
* Verify successful compilation
* Execute fast running acceptance tests


provider_test.go
```diff
@@ -409,7 +410,7 @@ func fromInstanceState(s *terraform.State, name, key string) (string, error) {
 		return "", fmt.Errorf("No primary instance: %s", name)
 	}
 
-	v, ok := is.Attributes[key];
+	v, ok := is.Attributes[key]
```
codycushing: Only actual code change from `goimports`

PR #267, opened 2017-09-25
Jlh instance chef

jcobol:
Add a Terraform + Chef example.  It is based on the compute/instance example.


docs/examples/compute/instance_chef/README.md
```diff
@@ -0,0 +1,58 @@
...
+    #   | | | | |_) | / _ \| |   | |   |  _|
+    #   | |_| |  _ < / ___ | |___| |___| |___
+    #    \___/|_| \_/_/   \_\____|_____|_____|
+***
+## Manage an instance
+This example launches an instance into an existing subnet, registers the instance with a Chef server, includes a user-data script in the instance launch, remote executes a command, and outputs the public and private IP address of the instance.
+
+### Prerequisites
+
+* Subscription to the Oracle Management Cloud Service  -https://cloud.oracle.com/management
```
briangustafson: Not required, right?

jcobol: Right - thanks for spotting that.

docs/examples/compute/instance_chef/README.md
```diff
@@ -0,0 +1,57 @@
+    #     ___  ____     _    ____ _     _____
+    #    / _ \|  _ \   / \  / ___| |   | ____|
+    #   | | | | |_) | / _ \| |   | |   |  _|
+    #   | |_| |  _ < / ___ | |___| |___| |___
+    #    \___/|_| \_/_/   \_\____|_____|_____|
+***
+## Manage an instance
+This example launches an instance into an existing subnet, registers the instance with a Chef server, includes a user-data script in the instance launch, remote executes a command, and outputs the public and private IP address of the instance.
+
+### Prerequisites
```
briangustafson: How berks as a prereq, along with the setup to communicate with your server. (Does berks come with Chef DK?)

Also, knife must be installed and configured and in your path in order for the "knife node delete" to work on destroy.

docs/examples/compute/instance_chef/compute.tf
```diff
@@ -0,0 +1,42 @@
+resource "oci_core_instance" "TFInstance" {
+  availability_domain = "${lookup(data.oci_identity_availability_domains.ADs.availability_domains[var.AD - 1],"name")}"
+  compartment_id = "${var.compartment_ocid}"
+  display_name = "TFInstance"
```
briangustafson: nit: How about ChefNodeExample ?

docs/examples/compute/instance_chef/datasources.tf
```diff
@@ -0,0 +1,23 @@
...
+# Gets the OCID of the OS image to use
+data "oci_core_images" "OLImageOCID" {
+    compartment_id = "${var.compartment_ocid}"
+    operating_system = "${var.InstanceOS}"
+    operating_system_version = "${var.InstanceOSVersion}"
+}
+
+# Gets a list of vNIC attachments on the instance
+data "oci_core_vnic_attachments" "InstanceVnics" {
+compartment_id = "${var.compartment_ocid}"
```
briangustafson: inconsistent formatting on this file

docs/examples/compute/instance_chef/variables.tf
```diff
@@ -0,0 +1,63 @@
...
+
+variable "InstanceOS" {
+    default = "Oracle Linux"
+}
+
+variable "InstanceOSVersion" {
+    default = "7.3"
+}
+
+variable "2TB" {
```
briangustafson: not used

docs/examples/compute/instance_chef/variables.tf
```diff
@@ -0,0 +1,63 @@
...
+    default = "2097152"
+}
+
+variable "256GB" {
```
briangustafson: not used

docs/examples/compute/instance_chef/cookbooks/example_webserver/recipes/default.rb
```diff
@@ -0,0 +1,30 @@
+#
+# Cookbook:: example_webserver
```
briangustafson: When I run this plan and go to the public IP in my browser, I get an Apache setup page. We could add some content of our own with something like the following (which I've tested, it works):

file '/var/www/html/index.html' do
  content '<html>Hello World!</html>'
end

docs/examples/compute/instance_chef/cookbooks/example_webserver/.kitchen.yml
```diff
@@ -0,0 +1,26 @@
+---
```
briangustafson: I would be in favor of leaving this and the test case out, just to streamline the example.

docs/examples/compute/instance_chef/block.tf
```diff
@@ -0,0 +1,14 @@
+resource "oci_core_volume" "TFBlock0" {
```
briangustafson: I would leave out the block volume part, since that doesn't really have anything to do wth the Chef example, and we show how to attach and mount a volume in other examples.

docs/examples/compute/instance_chef/README.md
```diff
@@ -0,0 +1,57 @@
...
+### Using this example
+* Update env-vars with the required information. Most examples use the same set of environment variables so you only need to do this once.
+* Source env-vars
+  * `$ . env-vars`
+* Update `variables.tf` with your instance options.
+* Upload the example Chef recipe and its dependencies to your Chef server.
+  * `$ cd cookbooks/example_webserver`
+  * `$ berks install`
+  * `$ berks upload`
+
```
briangustafson: Steps to add:
terraform apply
Go to the public IP address in your browser.

docs/examples/compute/instance_chef/README.md
jcobol: Berks is included with ChefDK according to https://docs.chef.io/berkshelf.html
I'll add a note about knife to the README.

docs/examples/compute/instance_chef/README.md
```diff
@@ -0,0 +1,57 @@
+    #     ___  ____     _    ____ _     _____
+    #    / _ \|  _ \   / \  / ___| |   | ____|
+    #   | | | | |_) | / _ \| |   | |   |  _|
+    #   | |_| |  _ < / ___ | |___| |___| |___
+    #    \___/|_| \_/_/   \_\____|_____|_____|
+***
+## Manage an instance
```
briangustafson: How about:

## Bootstrap an instance as a Chef node

This example shows how to use Terraform to bootstrap an OCI instance as a Chef node, register it with an existing Chef server, and run a recipe. When the plan completes, you will have a running web server with content and the correct firewall configuration.

PR #271, opened 2017-09-26
address possible race condition in policy test

scotmoor:
Separate out the Config "setup" steps from the verification to address a possible race condition that may be the cause of failures.


codycushing:
It's not a bullet proof way to address race conditions in test verification steps but it seems sufficient as far as I have tested

PR #275, opened 2017-09-28
fixing the policy format changes issue

parweza:
Testing Done:
manually created the policy resource and validated 
1. Apply does not update the policy even though format difference
2. server side policy update gets overwritten  on apply
[update.txt](https://github.com/oracle/terraform-provider-oci/files/1342432/update.txt)
[create.txt](https://github.com/oracle/terraform-provider-oci/files/1342433/create.txt)
[terraform.tfstate.backup.txt](https://github.com/oracle/terraform-provider-oci/files/1342442/terraform.tfstate.backup.txt)
[terraform.tfstate.txt](https://github.com/oracle/terraform-provider-oci/files/1342443/terraform.tfstate.txt)
[policy.tf.txt](https://github.com/oracle/terraform-provider-oci/files/1342444/policy.tf.txt)

### DIFF
--- a/resource_obmcs_identity_policy.go
+++ b/resource_obmcs_identity_policy.go
@@ -6,6 +6,10 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/oracle/bmcs-go-sdk"
 
+	"crypto/md5"
+	"encoding/hex"
+	"strings"
+
 	"github.com/oracle/terraform-provider-oci/crud"
 )
 
@@ -42,9 +46,22 @@ func PolicyResource() *schema.Resource {
 			Computed: true,
 		},
 		"statements": {
-			Type:     schema.TypeList,
-			Required: true,
-			Elem:     &schema.Schema{Type: schema.TypeString},
+			Type:             schema.TypeList,
+			Required:         true,
+			DiffSuppressFunc: ignorePolicyFormatDiff,
+			Elem:             &schema.Schema{Type: schema.TypeString},
+		},
+		"ETag": {
+			Type:     schema.TypeString,
+			Computed: true,
+		},
+		"policyHash": {
+			Type:     schema.TypeString,
+			Computed: true,
+		},
+		"lastUpdateETag": {
+			Type:     schema.TypeString,
+			Computed: true,
 		},
 		"inactive_state": {
 			Type:     schema.TypeInt,
@@ -67,6 +84,25 @@ func PolicyResource() *schema.Resource {
 		Delete:   deletePolicy,
 		Schema:   policySchema,
 	}
+}
+
+func ignorePolicyFormatDiff(k string, old string, new string, d *schema.ResourceData) bool {
+	oldHash := getOrDefault(d, "policyHash", "")
+	newHash := getMD5Hash(toStringArray(d.Get("statements")))
+	oldETag := getOrDefault(d, "lastUpdateETag", "")
+	currentETag := getOrDefault(d, "ETag", "")
+	suppressDiff := strings.EqualFold(oldHash, newHash) && strings.EqualFold(oldETag, currentETag)
+
+	return suppressDiff
+}
+
+func getOrDefault(d *schema.ResourceData, key string, defaultValue string) string {
+	valueString := defaultValue
+	if value, ok := d.GetOk(key); ok {
+		valueString = value.(string)
+	}
+
+	return valueString
 }
 
 func createPolicy(d *schema.ResourceData, m interface{}) (e error) {
@@ -131,7 +167,7 @@ func (s *PolicyResourceCrud) DeletedTarget() []string {
 	return []string{baremetal.ResourceDeleted}
 }
 
-func (s *PolicyResourceCrud) toStringArray(vals interface{}) []string {
+func toStringArray(vals interface{}) []string {
 	arr := vals.([]interface{})
 	result := []string{}
 	for _, val := range arr {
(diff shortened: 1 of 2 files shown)


resource_obmcs_identity_policy.go
```diff
@@ -165,23 +205,31 @@ func (s *PolicyResourceCrud) Update() (e error) {
...
 	s.D.Set("name", s.Res.Name)
 	s.D.Set("description", s.Res.Description)
 	s.D.Set("compartment_id", s.Res.CompartmentID)
 	s.D.Set("state", s.Res.State)
 	s.D.Set("time_created", s.Res.TimeCreated.String())
 }
 
+func getMD5Hash(values [] string) string {
+	statements := strings.Join(values, "#");
+	return fmt.Sprintf("%x", md5.Sum([]byte(statements)))
```
codycushing: Have a look at `hex.EncodeToString`

parweza: hex.EncodeToString  encodes the string into hexadecimal format, I want hash so that we can do equality comparison without having to storE the entire policy statement.

 Are you suggesting to do hex encoding before hashing or do you want me to store encoded string?

codycushing: You are using fmt.Sprintf to coerce a hex value to a string, I was suggesting using the intended utility for that conversion

codycushing:
What do you think about including this as an acceptance test case?

resource_obmcs_identity_policy.go
parweza: Oh, Ok. Done

parweza:
Updated Test cases

resource_obmcs_identity_policy_test.go
```diff
@@ -28,7 +28,7 @@ func (s *ResourceIdentityPolicyTestSuite) SetupTest() {
 	s.Config = testProviderConfig() + `
 	resource "oci_identity_compartment" "t" {
 		name = "-tf-compartment"
-		description = "tf test compartment"
+		description = "tf test compartment2"
```
codycushing: Is this change required?

resource_obmcs_identity_policy_test.go
```diff
@@ -67,6 +70,25 @@ func (s *ResourceIdentityPolicyTestSuite) TestAccResourceIdentityPolicy_basic()
...
+					statements = ["Allow group ${oci_identity_group.t.name} to read instances in    compartment ${oci_identity_compartment.t.name}"]
+				}`,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(s.ResourceName, "id"),
+					resource.TestCheckResourceAttrSet(s.ResourceName, "compartment_id"),
+					resource.TestCheckResourceAttrSet(s.ResourceName, "time_created"),
+					resource.TestCheckResourceAttr(s.ResourceName, "name", "-tf-policy"),
+					resource.TestCheckResourceAttr(s.ResourceName, "policyHash", "5f251343112c766d4ef6014f0c97c68c"),
+					resource.TestCheckResourceAttr(s.ResourceName, "description", "automated test policy"),
+					resource.TestCheckResourceAttr(s.ResourceName, "statements.#", "1"),
```
codycushing: I don't think we should duplicate assertions from previous steps for values that are not being changed, unless there is a reasonable suspicion that some change might have side effects.

resource_obmcs_identity_policy_test.go
```diff
@@ -51,12 +51,15 @@ func (s *ResourceIdentityPolicyTestSuite) TestAccResourceIdentityPolicy_basic()
 					compartment_id = "${oci_identity_compartment.t.id}"
 					name = "-tf-policy"
 					description = "automated test policy"
-					statements = ["Allow group ${oci_identity_group.t.name} to read instances in compartment ${oci_identity_compartment.t.name}"]
+					statements = ["Allow group ${oci_identity_group.t.name} to read instances in    compartment ${oci_identity_compartment.t.name}"]
```
codycushing: It might be clearer if this is tested in a dedicated update step, rather than implicitly in the creation step. However, since creation appears to be one of the areas this can fail, it could be clearer to split this into another test function, like TestAccResourceIdentityPolicy_postApply()

resource_obmcs_identity_policy_test.go
parweza: someone updated the description of compartment, this change was to make the test runnable while I fix the compartment description.
Fixed.

resource_obmcs_identity_policy_test.go
parweza: removed, kept only assertions that are validates that field values has not changed

resource_obmcs_identity_policy_test.go
parweza: update the tests as discussed

resource_obmcs_identity_policy.go
```diff
@@ -69,6 +85,27 @@ func PolicyResource() *schema.Resource {
...
+	return func(k string, old string, new string, d *schema.ResourceData) bool {
+		oldHash := getOrDefault(d, "policyHash", "")
+		newHash := getMD5Hash(toStringArray(d.Get("statements")))
+		oldETag := getOrDefault(d, "lastUpdateETag", "")
+		currentETag := getOrDefault(d, "ETag", "")
+		suppressDiff := strings.EqualFold(oldHash, newHash) && strings.EqualFold(oldETag, currentETag)
+
+		return suppressDiff
+	}
+}
```
codycushing: Why not just implement SchemaDiffSuppressFunc with the ignorePolicyFormatDiff function and pass it to the schema directly?

resource_obmcs_identity_policy.go
```diff
@@ -69,6 +85,27 @@ func PolicyResource() *schema.Resource {
...
+
+func getOrDefault(d *schema.ResourceData, key string, defaultValue string) string {
+	valueString := defaultValue
+	if value, ok := d.GetOk(key); ok {
+		valueString = value.(string)
+	}
+
+	return valueString;
```
codycushing: no semicolon

codycushing:
Be sure to run `make fmt` or install the pre-push hook

resource_obmcs_identity_policy.go
parweza: done

PR #276, opened 2017-09-29
Add object store fixes and file upload example

codycushing:
### Tests
ccushing-mac:terraform-provider-oci ccushing$ make test run=Test.*Objectstorage.*TestSuite
=== RUN   TestDatasourceObjectstorageBucketSummaryTestSuite
=== RUN   TestAccDatasourceObjectstorageBucketSummaries_basic
--- PASS: TestAccDatasourceObjectstorageBucketSummaries_basic (23.43s)
--- PASS: TestDatasourceObjectstorageBucketSummaryTestSuite (23.43s)
=== RUN   TestDatasourceObjectstorageNamespaceTestSuite
=== RUN   TestAccDatasourceObjectstorageNamespace_basic
--- PASS: TestAccDatasourceObjectstorageNamespace_basic (1.41s)
--- PASS: TestDatasourceObjectstorageNamespaceTestSuite (1.42s)
=== RUN   TestDatasourceObjectstorageObjectHeadTestSuite
=== RUN   TestObjectstorageObjectHead_basic
--- PASS: TestObjectstorageObjectHead_basic (23.75s)
--- PASS: TestDatasourceObjectstorageObjectHeadTestSuite (23.75s)
=== RUN   TestDatasourceObjectstorageObjectTestSuite
=== RUN   TestAccDatasourceObjectstorageObjects_basic
--- PASS: TestAccDatasourceObjectstorageObjects_basic (24.85s)
--- PASS: TestDatasourceObjectstorageObjectTestSuite (24.85s)
=== RUN   TestResourceObjectstorageBucketTestSuite
=== RUN   TestAccResourceObjectstorageBucket_basic
--- PASS: TestAccResourceObjectstorageBucket_basic (22.31s)
--- PASS: TestResourceObjectstorageBucketTestSuite (22.32s)
=== RUN   TestResourceObjectstorageObjectTestSuite
=== RUN   TestAccResourceObjectstorageObject_basic
--- PASS: TestAccResourceObjectstorageObject_basic (25.23s)
--- PASS: TestResourceObjectstorageObjectTestSuite (25.23s)
=== RUN   TestResourceObjectstoragePARTestSuite
=== RUN   TestAccResourceObjectstoragePAR_basic
--- PASS: TestAccResourceObjectstoragePAR_basic (23.39s)
--- PASS: TestResourceObjectstoragePARTestSuite (23.40s)
PASS
ok      github.com/oracle/terraform-provider-oci        144.405s

### DIFF
--- a/resource_obmcs_objectstorage_object.go
+++ b/resource_obmcs_objectstorage_object.go
@@ -10,7 +10,6 @@ import (
 
 	"crypto/md5"
 	"encoding/hex"
-	"strings"
 
 	"github.com/oracle/terraform-provider-oci/crud"
 )
@@ -40,7 +39,7 @@ func ObjectResource() *schema.Resource {
 				if v == "" {
 					return ""
 				}
-				h := md5.Sum([]byte(strings.TrimSpace(v)))
+				h := md5.Sum([]byte(v))
 				return hex.EncodeToString(h[:])
 			},
 		},
@@ -56,7 +55,7 @@ func ObjectResource() *schema.Resource {
 			Type:     schema.TypeInt,
 			Computed: true,
 		},
-		"content_MD5": {
+		"content_md5": {
 			Type:     schema.TypeString,
 			Computed: true,
 		},
@@ -131,7 +130,7 @@ func (s *ObjectResourceCrud) SetData() {
 	s.D.Set("content_encoding", s.Res.ContentEncoding)
 	s.D.Set("content_language", s.Res.ContentLanguage)
 	s.D.Set("content_length", s.Res.ContentLength)
-	s.D.Set("content_MD5", s.Res.ContentMD5)
+	s.D.Set("content_md5", s.Res.ContentMD5)
 	s.D.Set("content_type", s.Res.ContentType)
 }
 
--- a/resource_obmcs_objectstorage_object.go
+++ b/resource_obmcs_objectstorage_object.go
@@ -8,10 +8,69 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/oracle/bmcs-go-sdk"
 
+	"crypto/md5"
+	"encoding/hex"
+	"strings"
+
 	"github.com/oracle/terraform-provider-oci/crud"
 )
 
 func ObjectResource() *schema.Resource {
+	var objectSchema = map[string]*schema.Schema{
+		"namespace": {
+			Type:     schema.TypeString,
+			Required: true,
+			Computed: false,
+		},
+		"bucket": {
+			Type:     schema.TypeString,
+			Required: true,
+			Computed: false,
+		},
+		"object": {
+			Type:     schema.TypeString,
+			Required: true,
+			Computed: false,
+		},
+		"content": {
+			Type:     schema.TypeString,
+			Optional: true,
+			StateFunc: func(body interface{}) string {
+				v := body.(string)
+				if v == "" {
+					return ""
+				}
+				h := md5.Sum([]byte(strings.TrimSpace(v)))
+				return hex.EncodeToString(h[:])
+			},
+		},
+		"content_encoding": {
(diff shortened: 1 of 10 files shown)


resource_obmcs_objectstorage_object.go
```diff
@@ -8,10 +8,69 @@ import (
...
+		},
+		"content_language": {
+			Type:     schema.TypeString,
+			Optional: true,
+		},
+		"content_length": {
+			Type:     schema.TypeInt,
+			Computed: true,
+		},
+		"content_MD5": {
```
briangustafson: lower case content_md5

resource_obmcs_objectstorage_object.go
```diff
@@ -8,10 +8,69 @@ import (
...
+		},
+		"object": {
+			Type:     schema.TypeString,
+			Required: true,
+			Computed: false,
+		},
+		"content": {
+			Type:     schema.TypeString,
+			Optional: true,
+			StateFunc: func(body interface{}) string {
```
briangustafson: So this will no longer write the content to the state file, correct? Will this be an issue for users who upgrade?

resource_obmcs_objectstorage_object.go
```diff
@@ -8,10 +8,69 @@ import (
...
+				v := body.(string)
+				if v == "" {
+					return ""
+				}
+				h := md5.Sum([]byte(strings.TrimSpace(v)))
```
briangustafson: Should we trim space? I would think that the string is data that should stay intact.

resource_obmcs_objectstorage_object.go
codycushing: It will cause the same object to be put again

resource_obmcs_objectstorage_object.go
codycushing: agree

resource_obmcs_objectstorage_object.go
codycushing: done

vendor/github.com/oracle/bmcs-go-sdk/object_storage_objects.go
```diff
@@ -204,6 +204,15 @@ func (c *Client) PutObject(
 	}
 	required.Body = content
 
+	// application/json gets added by default if content-type is not specified. This is not desirable
```
briangustafson: So the service will set "application/octet-stream" if it doesn't get a content type, and we're just mirroring that here, correct?

resource_obmcs_objectstorage_object.go
briangustafson: Ok, sounds reasonable to have a one time update on upgrade, given that we don't really support large files.

resource_obmcs_objectstorage_object.go
```diff
@@ -94,6 +157,18 @@ func (s *ObjectResourceCrud) Update() (e error) {
 	content := s.D.Get("content").(string)
 	opts := &baremetal.PutObjectOptions{}
 
+	if contentEncoding, ok := s.D.GetOk("content_encoding"); ok {
```
briangustafson: Doesn't need to be for this PR, but why don't we just mark everything ForceNew and get rid of Update? I don't believe that objects have a real update - you must send the full set of data every time, and that overwrites what is there.

vendor/github.com/oracle/bmcs-go-sdk/object_storage_objects.go
codycushing: The sdk sets application/json if no content type is set which is correct/required for basically every other end point. For object store that is not a correct default and must have an explicit override of application/octet-stream (if none is supplied by the user). 

Ideally we could tell the sdk to not set a default content-type here but it is not sufficiently parameterized to support that.

PR #282, opened 2017-10-03
Document OCI migration when using remote state

briangustafson:


tools/oci-tool/readme.md
```diff
@@ -25,6 +25,12 @@ providers {
 
 Otherwise, drop the new provider in your terraform plugin directory.
 
+#### Remote State
+
+The **oci-tool** does not support [remote state](https://www.terraform.io/docs/state/remote.html). If you are using remote state, then the recommended approach is to copy the state file locally, run the **oci-tool**, and then push the stateful back to the remote location.
+
```
codycushing: stateful ->statefile

briangustafson: fixed

tools/oci-tool/readme.md
```diff
@@ -25,6 +25,12 @@ providers {
...
+If you remove the [backend configuration](https://www.terraform.io/docs/backends/config.html) and run `terraform init`, Terraform will ask if you'd like to copy the remote state file to a local locatation. After running **oci-tool** using the steps in the next section, adding the backend config back and running `terraform init` again will give you the option of uploading the updated state file.
+
```
codycushing: We should mention the version of Terraform to which this approach was employed, in case it behaves differently/unexpectedly for other versions

briangustafson: The documentation at https://www.terraform.io/docs/backends/config.html does back this up - but I agree that we should be cautious with this recommendation. For now I'm going to remove this paragraph.

codycushing: But it is edifying, I would just add a YMMV comment

briangustafson:
I went with a middle of the road approach in the latest commit. @codycushing Let me know if that makes sense.

codycushing:
looks good, just remember to squash when you merge

PR #284, opened 2017-10-04
Test fixes

briangustafson:
This fixes the following tests, which were all failing in the same way.
Note that we can probably improve the pattern, but for now I just
want to get these working by using the pattern that we have in most
other tests.

make test run=TestDatasourcePrivateIPTestSuite
make test run=ResourcePrivateIPTestSuite
make test run=DatabaseDBSystemShapeTestSuite
make test run=DatabaseDBVersionTestSuite
make test run=DatabaseDatabasesTestSuite
make test run=ResourceDatabaseDBSystemTestSuite

### DIFF
--- a/data_source_obmcs_core_private_ip_test.go
+++ b/data_source_obmcs_core_private_ip_test.go
@@ -6,7 +6,6 @@ import (
 	"testing"
 
 	"github.com/hashicorp/terraform/helper/resource"
-	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/oracle/bmcs-go-sdk"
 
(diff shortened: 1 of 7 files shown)


codycushing:
Does the ResourceDatabaseDBSystemTestSuite pass for you after this change? I have [a refactor branch](https://github.com/oracle/terraform-provider-oci/tree/cush-dbsystems-tests) with a superset of these changes and it failures on tear down because of an vnic attachment. 

Incidentally the test invocation you are using happens to work `make test run=ResourceDatabaseDBSystemTestSuite` because `ResourceDatabaseDBSystemTestSuite` is a substring of the runnable command `TestResourceDatabaseDBSystemTestSuite`

PR #291, opened 2017-10-06
version date formatting issue fix

parweza:
Manual and integ test


codycushing:
Should there be test updates and output in this PR?

parweza:
As discussed currently we dont support policy creation version date. I have filed a story to support policy creation with version date.

PR #301, opened 2017-10-11
Update and Refresh for create_vnic_details

briangustafson:
Schema TypeMap as is currently used for create_vnic_details does not support properties such
as Computed and ForceNew, which prevented using this with update and refresh. TypeList
does appear to handle these correctly, and switching does not require any changes from users.

Tests run:
make test run=ResourceCoreVnicAttachmentTestSuite debug=1
make test run=ResourceCoreInstanceTestSuite debug=1

### DIFF
--- a/helpers_core.go
+++ b/helpers_core.go
@@ -44,6 +44,7 @@ var createVnicDetailsSchema = &schema.Resource{
 	},
 }
 
+// vnicDetailsList is assumed to be non-nil and non-empty.
 func SetCreateVnicOptions(vnicDetailsList []interface{}) (vnicOpts *baremetal.CreateVnicOptions) {
 	vnic := vnicDetailsList[0].(map[string]interface{})
 
@@ -80,6 +81,7 @@ func SetCreateVnicOptions(vnicDetailsList []interface{}) (vnicOpts *baremetal.Cr
 	return
 }
 
+// vnicDetailsList is assumed to be non-nil and non-empty.
 func SetUpdateVnicOptions(vnicDetailsList []interface{}) (vnicOpts *baremetal.UpdateVnicOptions) {
 	vnic := vnicDetailsList[0].(map[string]interface{})
 	vnicOpts = &baremetal.UpdateVnicOptions{}
--- a/helpers_core.go
+++ b/helpers_core.go
@@ -3,13 +3,49 @@
 package main
 
 import (
-	"strconv"
-
+	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/oracle/bmcs-go-sdk"
 )
 
-func SetCreateVnicOptions(rawCreateVnicDetails interface{}) (vnicOpts *baremetal.CreateVnicOptions, err error) {
-	vnic := rawCreateVnicDetails.(map[string]interface{})
+var createVnicDetailsSchema = &schema.Resource{
+	Schema: map[string]*schema.Schema{
+		"assign_public_ip": {
+			Type:     schema.TypeBool,
+			Optional: true,
+			Default:  true,
+			ForceNew: true,
+		},
+		"display_name": {
+			Type:     schema.TypeString,
+			Optional: true,
+			Computed: true,
+		},
+		"hostname_label": {
+			Type:     schema.TypeString,
+			Optional: true,
+			Computed: true,
+		},
+		"private_ip": {
+			Type:     schema.TypeString,
+			Optional: true,
+			Computed: true,
+			ForceNew: true,
+		},
+		"skip_source_dest_check": {
+			Type:     schema.TypeBool,
+			Optional: true,
+			Default:  false,
+		},
+		"subnet_id": {
+			Type:     schema.TypeString,
+			Required: true,
+			ForceNew: true,
+		},
+	},
+}
+
+func SetCreateVnicOptions(vnicDetailsList []interface{}) (vnicOpts *baremetal.CreateVnicOptions) {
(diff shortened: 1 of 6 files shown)


briangustafson:
In addition to the regular test runs, I've done a bunch of testing on this to ensure that it won't cause issue for users with existing instances and secondary VNICs when they upgrade. For instances, I've verified the upgrade both with and without the use or create_vnic_details. It appears to work in every case.

helpers_core.go
```diff
@@ -29,21 +65,51 @@ func SetCreateVnicOptions(rawCreateVnicDetails interface{}) (vnicOpts *baremetal
...
+	if skipSourceDestCheck != nil {
+		vnicOpts.SkipSourceDestCheck = new(bool)
+		*vnicOpts.SkipSourceDestCheck = skipSourceDestCheck.(bool)
+	}
+
+	return
+}
+
+func SetUpdateVnicOptions(vnicDetailsList []interface{}) (vnicOpts *baremetal.UpdateVnicOptions) {
+	vnic := vnicDetailsList[0].(map[string]interface{})
```
codycushing: Do we want a len or nil check here?

resource_obmcs_core_instance_test.go
```diff
@@ -154,34 +161,109 @@ func (s *ResourceCoreInstanceTestSuite) TestAccResourceCoreInstance_basic() {
...
+				ImportState:       true,
+				ImportStateVerify: true,
 				Config: s.Config + `
 				resource "oci_core_instance" "t" {
 					availability_domain = "${data.oci_identity_availability_domains.ADs.availability_domains.0.name}"
 					compartment_id = "${var.compartment_id}"
 					image = "${data.oci_core_images.t.images.0.id}"
 					shape = "VM.Standard1.1"
 					display_name = "-tf-instance"
+					subnet_id = "${oci_core_subnet.t.id}"
```
codycushing: Does removing subnet_id here when adding the create_vnic_details object cause "force new"?

helpers_core.go
briangustafson: It *shouldn't* be an issue, since that is guaranteed before calling this: in VnicAttachment it's required, and in Instance we check with GetOk. And if create_vnic_details is present, then it has MinItems set to 1 (and MaxItems to 1), so it must have exactly 1 item. I'll add a comment to the method that this is an expectation of that parameter.

resource_obmcs_core_instance_test.go
```diff
@@ -154,34 +161,109 @@ func (s *ResourceCoreInstanceTestSuite) TestAccResourceCoreInstance_basic() {
...
+					func(ts *terraform.State) (err error) {
+						newId, err := fromInstanceState(ts, s.ResourceName, "id")
+						if newId != instanceId {
+							return fmt.Errorf("Expected same instance ocid, got different.")
+						}
+						return err
+					},
+				),
+			},
+			// verify force new by setting non-updateable VNIC details.
```
codycushing: updatable :#

resource_obmcs_core_instance_test.go
briangustafson: It does, and that's the current behavior without this change as well.

That said, I think we can change this pretty easily. Right now we're not reading in subnet_id in SetData. If we add it there and then mark it as calculated, then I believe users can switch off of this without consequence. I think there's enough in this PR already though, so I have not done that.

resource_obmcs_core_instance_test.go
```diff
@@ -154,34 +161,109 @@ func (s *ResourceCoreInstanceTestSuite) TestAccResourceCoreInstance_basic() {
...
+				Config: s.Config + `
+				resource "oci_core_instance" "t" {
+					availability_domain = "${data.oci_identity_availability_domains.ADs.availability_domains.0.name}"
+					compartment_id = "${var.compartment_id}"
+					image = "${data.oci_core_images.t.images.0.id}"
+					shape = "VM.Standard1.1"
+					display_name = "-tf-instance"
+					subnet_id = "${oci_core_subnet.t.id}"
+					metadata {
+						ssh_authorized_keys = "${var.ssh_public_key}"
```
codycushing: tangential not required change, we dont need to pass public keys to these test instances, can just pass `metadata {}`

codycushing: I guess we are generally testing typical params of an instance though so 
ignoreme

resource_obmcs_core_vnic_attachment_test.go
```diff
@@ -72,6 +75,55 @@ func (s *ResourceCoreVnicAttachmentTestSuite) TestAccResourceCoreVnicAttachment_
...
+					func(ts *terraform.State) (err error) {
+						vaId, err = fromInstanceState(ts, s.ResourceName, "id")
+						return err
+					},
+				),
+			},
+			{
+				// Update the VNIC
+				ImportState:       true,
+				ImportStateVerify: true,
```
codycushing: This test looks to be doing `ImportState`, `ImportStateVerify` in each step, perhaps not intentionally.

resource_obmcs_core_instance_test.go
briangustafson: https://english.stackexchange.com/questions/56431/updatable-vs-updateable-which-is-correct

Looks like they're both valid (although updatable is more popular).

resource_obmcs_core_vnic_attachment_test.go
briangustafson: removed all but 2 of these

PR #303, opened 2017-10-11
Test consistency pass

codycushing:
* More general subnet compartment_id assertion
* Fix inconsistent test related naming
* Convert resource.UnitTest to resource.Test in some missed locations
** There are a few more in subsequent commits


resource_obmcs_core_subnet_test.go
```diff
@@ -57,7 +57,7 @@ func TestAccResourceCoreSubnetCreate_basic(t *testing.T) {
 					resource.TestCheckResourceAttrSet(resourceName, "time_created"),
 					resource.TestCheckResourceAttrSet(resourceName, "virtual_router_ip"),
 					resource.TestCheckResourceAttrSet(resourceName, "virtual_router_mac"),
-					resource.TestMatchResourceAttr(resourceName, "compartment_id", regexp.MustCompile("ocid1\\.compartment\\.oc1\\..*")),
```
codycushing: This is typically "compartment" but could also be "tenancy", remove explicit check since we trust the server is returning the right thing, just verify the propery exists.

PR #304, opened 2017-10-11
Optimize Private Ip tests

codycushing:
* These tests were taking ~23 minutes, consolidated steps to get ~6 minutes
* Begin test_helpers for systematic approach to test fixture definitions
* Test Provider definition that properly reads in env variables

### TEST
=== RUN   TestResourceCorePrivateIPTestSuite
=== RUN   TestAccCoreResourcePrivateIP_basic
--- PASS: TestAccCoreResourcePrivateIP_basic (209.98s)
--- PASS: TestResourceCorePrivateIPTestSuite (209.98s)
=== RUN   TestDatasourceCorePrivateIPTestSuite
=== RUN   TestAccCorePrivateIPs_basic
--- PASS: TestAccCorePrivateIPs_basic (217.08s)
--- PASS: TestDatasourceCorePrivateIPTestSuite (217.08s)

### DIFF
--- a/data_source_obmcs_core_private_ip_test.go
+++ b/data_source_obmcs_core_private_ip_test.go
@@ -19,26 +19,29 @@ type DatasourcePrivateIPTestSuite struct {
 	Provider     terraform.ResourceProvider
 	Providers    map[string]terraform.ResourceProvider
 	ResourceName string
-	List         *baremetal.ListPrivateIPs
 }
 
 func (s *DatasourcePrivateIPTestSuite) SetupTest() {
 	s.Client = testAccClient
 	s.Provider = testAccProvider
 	s.Providers = testAccProviders
-	s.Config = vnicConfig + `
-resource "oci_core_private_ip" "testPrivateIP" {
-	vnic_id = "${lookup(data.oci_core_vnic_attachments.vnics.vnic_attachments[0],"vnic_id")}"
-	ip_address = "10.0.1.23"
-}
-	`
-	s.Config += testProviderConfig()
-	s.ResourceName = "data.oci_core_private_ips.testPrivateIPs"
+	s.Config = testProvider1() + testADs() + testVCN1() + testSubnet1() + testImage1() + testInstance1() + `	
+	data "oci_core_vnic_attachments" "t" {
+		compartment_id = "${var.compartment_ocid}"
+		availability_domain = "${data.oci_identity_availability_domains.t.availability_domains.0.name}"
+		instance_id = "${oci_core_instance.t.id}"
+	}
+	
+	resource "oci_core_private_ip" "t" {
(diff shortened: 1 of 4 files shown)


resource_obmcs_core_private_ip_test.go
```diff
@@ -17,134 +18,100 @@ type ResourcePrivateIPTestSuite struct {
...
+					resource.TestCheckResourceAttrSet(s.ResourceName, "vnic_id"),
+					resource.TestCheckResourceAttrSet(s.ResourceName, "ip_address"),
+					resource.TestCheckResourceAttrSet(s.ResourceName, "availability_domain"),
+					resource.TestCheckResourceAttrSet(s.ResourceName, "compartment_id"),
+					resource.TestCheckResourceAttrSet(s.ResourceName, "is_primary"),
+					resource.TestCheckResourceAttrSet(s.ResourceName, "subnet_id"),
+					resource.TestCheckResourceAttrSet(s.ResourceName, "time_created"),
+					resource.TestCheckResourceAttr(s.ResourceName, "display_name", "-private-ip"),
+					func(s *terraform.State) (err error) {
+						resId, err = fromInstanceState(s, "oci_core_private_ip.t", "id")
```
briangustafson: Looks like resId is checked for the final destructive change, but might as well verify for the updates as well. No blocking.

data_source_obmcs_core_private_ip_test.go
```diff
@@ -19,26 +19,29 @@ type DatasourcePrivateIPTestSuite struct {
...
-	s.Config = vnicConfig + `
-resource "oci_core_private_ip" "testPrivateIP" {
-	vnic_id = "${lookup(data.oci_core_vnic_attachments.vnics.vnic_attachments[0],"vnic_id")}"
-	ip_address = "10.0.1.23"
-}
-	`
-	s.Config += testProviderConfig()
-	s.ResourceName = "data.oci_core_private_ips.testPrivateIPs"
-}
+	s.Config = testProvider1() + testADs() + testVCN1() + testSubnet1() + testImage1() + testInstance1() + `
```
briangustafson: neat... would it make sense for each test resource to include its dependencies?

codycushing: That model gets complicated and breaks down pretty quickly when you need two things that both require the same upstream dependencies. I think aggregating functions that use the above could be a reasonable balance, example:
```
func testSubnetScenario() string {
  return testProvider1() + testADs() + testVCN1() + testSubnet1()
}

func testInstanceScenario() string {
  return testSubnetScenario() + testImage1() + testInstance1() 
}
```
But this can lead to its own complexity, packaged differently--what would the testInstanceScenario() look like if it requires 2 instances and 2 subnets?

Long term this is the beginning of being able to parameterize some of these functions.

PR #305, opened 2017-10-11
Refactor dbsystems tests

codycushing:
* Convention based naming
* resource.UnitTest -> resource.Test
* Further abandon use of provider_test.go as a location for all test fixture defs

Note: resource test fails during destroy phase because subnet is destroyed before vnic is actually detached, investigating further if resource issue, should not block landing actual test refinements though


data_source_obmcs_database_databases_test.go
```diff
@@ -41,38 +83,36 @@ func (s *DatabaseDatabasesTestSuite) TestReadDatabases() {
...
 				Config: s.Config + `
 				data "oci_database_db_systems" "t" {
-				  compartment_id = "${var.compartment_id}"
+					compartment_id = "${var.compartment_id}"
 				}
 				data "oci_database_db_homes" "t" {
-				  compartment_id = "${var.compartment_id}"
-				  db_system_id = "${oci_database_db_system.t.id}"
+					compartment_id = "${var.compartment_id}"
+					db_system_id = "ocid1.dbsystem.oc1.phx.abyhqljskxaehm3osd7ekh5sxqdp7y6ob3u54ayo6y57pcvcbk3cf75kd2oa"
```
scotmoor: Will this fail if the tests are configured to use a different tenancy and/or different region? Should it be set in an env variable?

data_source_obmcs_database_databases_test.go
```diff
@@ -41,38 +83,36 @@ func (s *DatabaseDatabasesTestSuite) TestReadDatabases() {
...
+					  database_id = "${data.oci_database_databases.t.databases.0.id}"
 				}
-			        data "oci_database_db_nodes" "t" {
-				      compartment_id = "${var.compartment_id}"
-				      db_system_id = "${oci_database_db_system.t.id}"
-			        }
-				`,
+				data "oci_database_db_nodes" "t" {
+					compartment_id = "${var.compartment_id}"
+					db_system_id = "ocid1.dbsystem.oc1.phx.abyhqljskxaehm3osd7ekh5sxqdp7y6ob3u54ayo6y57pcvcbk3cf75kd2oa"
```
scotmoor: same as above

data_source_obmcs_database_databases_test.go
codycushing: Good catch, I was running these tests against an existing DBSystem to speed things along, will update

data_source_obmcs_database_databases_test.go
```diff
@@ -11,25 +11,67 @@ import (
...
 	s.Provider = testAccProvider
 	s.Providers = testAccProviders
-	s.Config = databaseConfig
+	s.Config = testProviderConfig() + `
+	data "oci_identity_availability_domains" "ADs" {
+		compartment_id = "${var.compartment_id}"
+	}
 
-	s.Config += testProviderConfig()
+	resource "oci_core_virtual_network" "t" {
```
parweza: can we name the resource something meaningful instead t, just like ADs

codycushing: These are fairly consistently named "t" in the acceptance tests, if we want to go with a new convention we should plan for it, especially in the context of code generating the tests. These names aren't particularly meaningful in the test context anyway since you typically don't create more than one of a thing, and references require the complete resource name prefix already.

PR #309, opened 2017-10-12
Avoid name collisions with concurrent tests

codycushing:
* Helpers + test integration for creating unique, repeatable values to inject in template strings
* There are a few more integration points (remaining identity tests + object store) that we can integrate if this approach is acceptable

Tests: 
make test run=TestResourceIdentity

### DIFF
--- a/resource_obmcs_identity_user_test.go
+++ b/resource_obmcs_identity_user_test.go
@@ -31,6 +31,7 @@ func (s *ResourceIdentityUserTestSuite) SetupTest() {
 }
 
 func (s *ResourceIdentityUserTestSuite) TestAccResourceIdentityUser_basic() {
+	token, tokenFn := tokenize()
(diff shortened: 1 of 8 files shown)


resource_obmcs_identity_user_test.go
```diff
@@ -31,32 +31,36 @@ func (s *ResourceIdentityUserTestSuite) SetupTest() {
...
-				resource "oci_identity_user" "t" {
-					name = "-tf-user"
-					description = "automated test user (updated)"
-				}`,
+				Config: s.Config + tokenFn(
+					`resource "oci_identity_user" "t" {
+						name = "{{.token}}"
+						description = "{{.description}}"
+					}`,
+					map[string]string{"description": "automated test user (updated)"}),
```
codycushing: Example of regular template string functionality being exposed

PR #316, opened 2017-10-16
Sync instance subnet and hostname_label

briangustafson:
An instances have 2 ways to set subnet_id and hostname_label: either at the
root level, or in create_vnic_details. By syncing both sets of properties with
the VNIC and setting all of these to computed, this change allows users to
switch between which version is used without causing changes to their plan.
This was partly in place before, but this change puts a few missing pieces in
place.
Note that hostname_label is updatable if specified in create_vnic_details, but not when
specified directly in the instance. I think that's fine for now - we could either add logic 
to work around that, or - what I would prefer - we can deprecate the instance version of
hostname_label (and subnet_id).

### DIFF
--- a/resource_obmcs_core_instance_test.go
+++ b/resource_obmcs_core_instance_test.go
@@ -83,6 +83,7 @@ func (s *ResourceCoreInstanceTestSuite) TestAccResourceCoreInstance_basic() {
 					availability_domain = "${data.oci_identity_availability_domains.ADs.availability_domains.0.name}"
 					compartment_id = "${var.compartment_id}"
 					subnet_id = "${oci_core_subnet.t.id}"
+					hostname_label = "hostname1"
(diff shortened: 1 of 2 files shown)


resource_obmcs_core_instance_test.go
```diff
@@ -105,7 +106,31 @@ func (s *ResourceCoreInstanceTestSuite) TestAccResourceCoreInstance_basic() {
...
+					image = "${data.oci_core_images.t.images.0.id}"
+					shape = "VM.Standard1.1"
+					metadata {
+						ssh_authorized_keys = "${var.ssh_public_key}"
+					}
+					timeouts {
+						create = "15m"
+					}
+				}`,
+				ExpectNonEmptyPlan: false,
```
codycushing: Comically one of the least intuitive boolean names I have come across

PR #317, opened 2017-10-19
Add VMDB and licenseModel support for DBSystem

kohashim:
Add the VMDB related attributes and "licenseModel" attribute to DBSystem to support VMDB and BYOL.

Currently running "	ResourceDatabaseDBSystemTestSuite".  Once it's complete, will update here.

Will update the changes under /vendor once the corresponding PR is approved and pushed.


resource_obmcs_database_db_system.go
```diff
@@ -378,9 +409,12 @@ func (s *DBSystemResourceCrud) SetData() {
 	s.D.Set("backup_subnet_id", s.Res.BackupSubnetID)
 	s.D.Set("cluster_name", s.Res.ClusterName)
 	s.D.Set("data_storage_percentage", s.Res.DataStoragePercentage)
+	s.D.Set("data_storage_size_in_gb", s.Res.DataStorageSizeInGB)
```
briangustafson: I don't see data_storage_size_in_gb in the schema. It looks like LaunchDbSystemDetails uses initialDataStorageSizeInGB, but DbSystem uses dataStorageSizeInGB. For the provider we should combine these into dataStorageSizeInGB.

data_source_obmcs_database_db_system.go
```diff
@@ -101,15 +101,19 @@ func (s *DBSystemDatasourceCrud) SetData() {
...
 				"display_name":            r.DisplayName,
 				"disk_redundancy":         r.DiskRedundancy,
 				"domain":                  r.Domain,
 				"hostname":                r.Hostname,
 				"id":                      r.ID,
+				"license_model":           r.LicenseModel,
 				"lifecycle_details":       r.LifecycleDetails,
 				"listener_port":           int(r.ListenerPort),
+				"node_count":              int(r.NodeCount),
+				"reco_storage_size_in_gb": int(r.RecoStorageSizeInGB),
```
briangustafson: Is this in the resource as well?

kohashim: "node_count" is in the resource, but "reco_storage_size_in_gb" is not.  According to their API spec, "reco_storage_size_in_gb" looks like a derived (computed) attribute, i.e, the request doesn't contain the attribute.

kohashim: Excuse me, I now understand what you mean behind the question.  Will add "reco_storage_size_in_gb" to the schema.  Thanks.

resource_obmcs_database_db_system.go
kohashim: Excuse me, I'm still learning how to write a provider.  Is the schema shared between Resource and DataSource?  Anyway, it makes sense to combine and unify the attribute name.  I will still leave the data_storage_size_in_gb as ForceNew=true for now because we don't support UpdateDbSystem yet.

kohashim:
"ResourceDatabaseDBSystemTestSuite" is succeeding now.

The public document https://docs.us-phoenix-1.oraclecloud.com/api/#/en/database/20160918/requests/DbSystem is inaccurate:
"dataStorageSizeInGB" -> "dataStorageSizeInGBs"

resource_obmcs_database_db_system_test.go
```diff
@@ -77,10 +54,37 @@ func (s *ResourceDatabaseDBSystemTestSuite) TestAccResourceDatabaseDBSystem() {
...
+						database {
+							"admin_password" = "BEstrO0ng_#11"
+							"db_name" = "aTFdb"
+							character_set = "AL32UTF8"
+							ncharacter_set = "AL16UTF16"
+						}
+					}
+				}`,
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttrSet(s.ResourceName, "id"),
```
briangustafson: Could definitely use some more validation on this test - but that can be done separately.

PR #326, opened 2017-10-24
Add solutions folder

briangustafson:
The docs/examples folder currently contains a mix of low-level examples and
very high level "how to complete scenario X" examples. This change starts the
migration of that second cagetory into a new section under docs/solutions.
(There are a few other examples that should probably be moved as well, but I need
to get signoff from the owners first.)
This also adds a placeholder for an upcoming SQL Server solution.
This puts the following into solutions:
- chef
- omc_monitored_server
- sql_server_always_on_availability_groups


codycushing:
If someone has executed one of these solutions from the existing folder structure and it contains their statefile and/or modifications, then they pull this change, will it have a destructive impact for them or leave their existing files in place? My guess is they will have a merge conflict when pulling--maybe we should give a deprecation notice for the existing examples structure before landing this and/or provide migration instructions.

briangustafson:
@codycushing Good question... I'm pretty sure that if you have a statefile in one of these examples that is being moved, that the statefiles will remain in its old location without any sort of conflict. Statefiles are part of .gitignore. Users would then need to copy that statefile to the new location in order to continue using it. (I'll do a bit of testing to make sure that this is correct.)

briangustafson:
Verified that my previous statement is correct: If you have an existing statefile for one of the moved examples, then you can simply copy it to the new location after syncing to this change and everything will continue to work as normal.

PR #329, opened 2017-10-26
Integrate filters in for Core Data Sources

codycushing:
* branched off preceding filter PR so includes previous filter.go and filter_test.go code
* currently does not include ipsec* or private ip datasources


data_source_obmcs_core_drg_test.go
```diff
@@ -19,41 +19,52 @@ type DatasourceCoreDrgTestSuite struct {
...
 	s.ResourceName = "data.oci_core_drgs.t"
 }
 
 func (s *DatasourceCoreDrgTestSuite) TestAccDatasourceCoreDrg_basic() {
-
 	resource.Test(s.T(), resource.TestCase{
 		PreventPostDestroyRefresh: true,
 		Providers:                 s.Providers,
 		Steps: []resource.TestStep{
+			//{
```
briangustafson: remove?

data_source_obmcs_core_drg_test.go
```diff
@@ -19,41 +19,52 @@ type DatasourceCoreDrgTestSuite struct {
...
 				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(s.ResourceName, "drgs.#", "1"),
+					resource.TestCheckResourceAttr(s.ResourceName, "drgs.0.display_name", s.Token),
+					resource.TestCheckResourceAttr(s.ResourceName, "drgs.0.state", "AVAILABLE"),
 					resource.TestCheckResourceAttrSet(s.ResourceName, "drgs.0.id"),
-					resource.TestCheckResourceAttrSet(s.ResourceName, "drgs.#"),
+					resource.TestCheckResourceAttrSet(s.ResourceName, "drgs.0.time_created"),
+					resource.TestCheckResourceAttrSet(s.ResourceName, "drgs.0.compartment_id"),
 				),
+				ExpectNonEmptyPlan: true,
```
briangustafson: does this check make sense?

data_source_obmcs_core_instances_test.go
```diff
@@ -84,15 +84,15 @@ func (s *DatasourceCoreInstanceTestSuite) TestAccDatasourceCoreInstance_basic()
...
-				Config:            s.Config,
-			},
-			{
 				Config: s.Config + s.TokenFn(`
 				data "oci_core_instances" "t" {
 					compartment_id = "${var.compartment_id}"
 					availability_domain = "${data.oci_identity_availability_domains.ADs.availability_domains.0.name}"
 					display_name = "{{.token}}"
-					limit = 1
+					filter {
```
briangustafson: As discussed, would be good to see what happens with metadata/extended_metadata.

data_source_obmcs_core_drg_test.go
codycushing: That was a workaround for a terraform depends_on bug with datasources, but I figured a way to define dependencies that obviates this so removed.

data_source_obmcs_core_instances_test.go
codycushing: Attempting to filter on maps or lists of structured objects returns no results. Maybe thats not ideal behavior but it's not a realistic use case until tags are integrated anyway.

PR #332, opened 2017-10-26
Add write up on filter usage

codycushing:


docs/Filters.md
```diff
@@ -0,0 +1,62 @@
...
+
+```
+filter {
+	name = ""
+	values = [""]
+}
+```
+
+The `name` value corresponds to the property name to filter against 
+and the `values` lists can contain one or more values filter with.
```
briangustafson: "values to filter with"

docs/Filters.md
```diff
@@ -0,0 +1,62 @@
...
+	name = ""
+	values = [""]
+}
+```
+
+The `name` value corresponds to the property name to filter against 
+and the `values` lists can contain one or more values filter with. 
+
+Multiple `values` work as an `OR` type filter. In the shape 
+example below the resulting data source would contain both VM
```
briangustafson: comma after below

PR #339, opened 2017-10-27
Create RHEL 7.4 bootable image

nelsonse:
This provides a fully automated process to create a RHEL 7.4 image in OCI.  Consult the README for requirements.


docs/solutions/rhel74_image/settfenv.py
```diff
@@ -0,0 +1,70 @@
...
+argparser.add_argument('--public_key', required=False,
+                       default="~/.ssh/id_rsa.pub", dest='public_key')
+args = vars(argparser.parse_args())
+
+outputEnv = {}
+outputFile = os.open(args["dest_file"], os.O_RDWR |
+                     os.O_TRUNC | os.O_CREAT, 0755)
+
+if len(args["config_path"]) <= 0:
+    obmcsConfig = "~/.oraclebmc/config"
```
briangustafson: ~/.oci/config

docs/solutions/rhel74_image/ipxe.sh.template
```diff
@@ -0,0 +1,167 @@
...
+boot
+EOF
+
+cd /var/www/html/oci
+cloud
+firewallcfg
+
+cat > /home/opc/sacrifice-launch.sh.template <<-EOF
+#!/bin/bash
+sac_inst_ocid=\`oci compute instance launch --availability-domain=${AD} \
```
briangustafson: This should all be much easier to do using the SDK instead of the CLI.

docs/solutions/rhel74_image/README.md
```diff
@@ -0,0 +1,40 @@
+This example provides method to generate a RHEL 7.4 image for use by both VM and BM shapes.
+There are several prerequisites for using this process:
+
```
briangustafson: If there's a relevant white paper to go with this, would make sense to link to it.

briangustafson:
Adding @dr-wesley to review the README.

docs/solutions/rhel74_image/README.md
```diff
@@ -0,0 +1,40 @@
+This example provides method to generate a RHEL 7.4 image for use by both VM and BM shapes.
```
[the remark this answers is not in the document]
nelsonse: Done

docs/solutions/rhel74_image/ipxe.sh.template
```diff
@@ -0,0 +1,167 @@
...
+
+systemctl enable httpd
+systemctl start httpd
+
+mkdir /var/www/isos
+mkdir /mnt/iso
+mkdir /var/www/html/ipxe
+mkdir /var/www/html/oci
+mkdir /var/www/html/<OS_NAME>
+mkdir -p /home/opc/.oraclebmc
```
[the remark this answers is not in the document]
nelsonse: done

docs/solutions/rhel74_image/ipxe.sh.template
nelsonse: Yes...as discussed.

docs/solutions/rhel74_image/ipxe.sh.template
```diff
@@ -0,0 +1,167 @@
...
+
+chmod 755 /home/opc/.oraclebmc
+chmod 777 /var/www/isos
+
+easy_install pip
+pip install pip --upgrade
+pip install oci
```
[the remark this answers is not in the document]
nelsonse: removed explicit oci install via pip

docs/solutions/rhel74_image/settfenv.py
nelsonse: Done

docs/solutions/rhel74_image/datasources.tf
```diff
@@ -0,0 +1,59 @@
...
+	compartment_id = "${data.oci_identity_compartments.compartment.compartments.0.id}"
+	vcn_id = "${data.oci_core_virtual_networks.vcn.virtual_networks.0.id}"
+	filter {
+		name = "display_name"
+		values = [ "${var.build_env["subnet"]}" ]
+	}
+
+}
+
+data "oci_core_images" "image" {
```
briangustafson: This method of getting the image OCID is used in most of our examples, but it will be changing in the near future. For now, we have the following comment on each usage - please add this comment here as well:

# Gets the OCID of the image. This technique is for example purposes only. The results of oci_core_images may
# change over time for Oracle-provided images, so the only sure way to get the correct OCID is to supply it directly.

docs/solutions/rhel74_image/ipxe.sh.template
```diff
@@ -0,0 +1,161 @@
...
+while [ \${ping_result} == 0 ]
+do
+	sleep 10
+	ping -q -c 1 \${sac_ip}
+	ping_result=\$?
+done
+image_ocid=\`oci compute image create --compartment-id=${COMPARTMENT} --instance-id=\${sac_inst_ocid} \
+--display-name="RHEL_74" | jq -r '.data["id"]'\`
+image_status=\`oci image get --image-id=\${image_ocid} | jq -r '.data["lifecycle-state"]'\`
+while [ "\${image_status}" != "AVAILABLE" ]
```
briangustafson: If something goes wrong with the image creation, this will be an infinite loop - same goes for the other loops in this script. Not blocking, but would be good put in a timeout. (Or is there some higher level timeout that could be applied to the entire ipxe script?)

docs/solutions/rhel74_image/settfenv.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/local/bin/python3.6
```
briangustafson: Add comment to top with instructions and example command to run this.

docs/solutions/rhel74_image/settfenv.py
```diff
@@ -0,0 +1,69 @@
...
+                      os.O_TRUNC | os.O_CREAT, 0x0755)
+
+if len(args["config_path"]) <= 0:
+    oci_config_file = "~/.oci/config"
+else:
+    oci_config_file = args["config_path"]
+
+oci_config_file = os.path.expanduser(oci_config_file)
+
+parser = configparser.ConfigParser()
```
briangustafson: It would be much simpler to use the OCI Python SDK for this, which already has a class for parsing and the config file. See "config = oci.config.from_file" in the example at https://oracle-cloud-infrastructure-python-sdk.readthedocs.io/en/latest/.

docs/solutions/rhel74_image/settfenv.py
briangustafson: I'd be in favor of leaving settfenv.py out for now. It's not necessary to this solution, or specific to it. It might make sense to give users a utility like this, but I believe it should be in a different location.

docs/solutions/rhel74_image/ipxe.sh.template
```diff
@@ -0,0 +1,161 @@
...
+EOF
+else
+	cat > config <<-EOF
+[DEFAULT]
+user=${USER}
+fingerprint=${FINGERPRINT}
+key_file=/home/opc/.oci/oci_api_key.pem
+tenancy=${TENANCY}
+region=${REGION}
+pass_phrase=${PR_KEY_PW}
```
briangustafson: Not 100% sure, but I believe that if there is no passphrase then you can just set "pass_phrase=" and it will be fine. So if you're looking to reduce the character count, that's something you can try.

docs/solutions/rhel74_image/ipxe.sh.template
nelsonse: How long would you suggest?  The OS install time is indeterminate, so a good run may be interrupted and there is no way to deterministically detect completion.

docs/solutions/rhel74_image/ipxe.sh.template
nelsonse: Yes, but if a pass_phrase is set, I need this in the config file that is generated as part of this.  This segment generates two different config files based on whether the passphrase is set.  I have tried setting the pass_phrase to null in the past and received errors.  If this behavior has changed I am unaware.

docs/solutions/rhel74_image/settfenv.py
nelsonse: Fine.  Deleted

docs/solutions/rhel74_image/ipxe.sh.template
briangustafson: I don't have the necessarily data to put a number on this. I do know that image creation can take quite a while. 

Another option (that can be done in addition to or instead of) is to look for known failure states, such as if the lifecycle state goes to DELETED.

Again, nice to have some safety checks here, but should not be blocking.

nelsonse: If there were the possibility of a safety check, I would have put it in.  Since the install process is completely blind, the only possible states of the system are AVAILABLE and PROVISIONING. I will put in a DELETED in the middle of the loop to allow for a manual destruction process.

docs/solutions/rhel74_image/README.md
```diff
@@ -0,0 +1,43 @@
...
+There are several prerequisites for using this process:
+
+1. You MUST have a valid RedHat account with subscriptions available.  The TF template needs a 
+   RH Username and Password to allow you to temporarily subscribe the instance that is building the image 
+   and get access to the various RH repos.
+2. The template expects pre-configured VCNs and Subnets.  
+3. You need to have a bucket in the OCI object store that contains the RHEL 7.4 iso.  The bucket must be in 
+   your tenancy, and must be accessible by the user executing the build.
+4. The template uses filters that expect unique Compartment, VCN and Subnet names.
+5. The following must be specified in your shell environment (prefixed with TF_VAR_ of course:
```
briangustafson: Not blocking, but most of our examples/solutions provide an env-vars file for setting all necessary variables. (That could also include the info being set in configuration.tf.)

docs/solutions/rhel74_image/README.md
```diff
@@ -0,0 +1,43 @@
...
+    - tenancy_ocid
+    - user_ocid
+    - fingerprint
+    - private_key_path
+    - private_key_password (if required)
+    - ssh_public_key (the actual public key, not the file)
+    - region
+    
+    There is a tool included in this directory, settfenv.py, that will generate a file that can be sourced into
```
briangustafson: This script is not longer in the PR.

docs/solutions/rhel74_image/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+What happens in the background:
+The template generates a script that embeds all the configuration files needed to build the iPXE server, extract the ISO
+boot the instance used to load RHEL, causes RHEL to load, builds the image, destroys the build instance, and finally destroys the iPXE server.  You are left with a custom image named "RHEL_74" in your environment.
+
+NOTE: The source configuration files for the iPXE server are included here.  It is *STRONGLY* recommended that they not be 
+      altered.
+      
+      
+ALSO NOTE: *THE PRIVATE KEY USED TO ACCESS OCI WILL TEMPORARILY BE TRANSFERRED TO THE IPXE INSTANCE.  ONCE THE IPXE INSTANCE IS DESTROYED, THE COPY OF THE PRIVATE KEY IS DESTROYED ALONG WITH IT.  THE PRIVATE KEY IS TRANSFERED IN A UUENCODED FORMAT TO OBSCURE THE KEY.*
```
briangustafson: As discussed, should remove the line "THE PRIVATE KEY IS TRANSFERED IN A UUENCODED FORMAT TO OBSCURE THE KEY."

docs/solutions/rhel74_image/README.md
briangustafson: Missing closing ) in "(prefixed with TF_VAR_ of course:".

nelsonse: Done

docs/solutions/rhel74_image/README.md
nelsonse: Done - removed reference

docs/solutions/rhel74_image/README.md
nelsonse: Done

PR #341, opened 2017-10-31
Update example image version to Oracle Linux 7.4

briangustafson:
Many of our examples and tests default to using Oracle Linux 7.3.
While 7.3 images can still be used if you know the OCID, they are no longer
returned by ListImages. Updating to 7.4 for now, but this is a short term
solution while we determine the correct way to handle updates to ListImages.


briangustafson:
Agree - also opened #342 to give this visibility.

Issue #345, opened 2017-11-01
oci_core_images returns incompatible GPU image as the first available Oracle Linux 7.4


jlamillan:
### Description:

It looks like oci_core_images is now returning a GPU image as the first image in the list, which will cause an error _unless_ the user is using the BM.GPU2.2 shape.

IMO, this is a problem since a lot of the example code sets up their oci_core_images data source like:

```
data "oci_core_images" "ImageOCID" {
  compartment_id           = "${var.compartment_ocid}"
  operating_system         = "${var.os}"
  operating_system_version = "${var.version}"
}
```

and use it like:

```
resource "oci_core_instance" "TFInstance" {
  ...
  image               = "${lookup(data.oci_core_images.ImageOCID.images[0], "id")}"
  shape               = ANY SHAPE BESIDES BM.GPU2.2
  ...
}
```

However, when `var.os=Oracle Linux` and `var.version=7.4` and you try and create an instance, you'll get an error:

### Terraform Plan

<details>
```
* oci_core_instance.TFInstanceK8sMaster: Status: 400; Code: InvalidParameter; OPC Request ID: /XXXXX; Message: Shape VM.Standard1.8 is not valid for image ocid1.image.oc1.iad.aaaaaaaaflaxipg2k4l7mjudowij5vic76jjnqgzktw46pga4knjpqcrxw6a.
```
</details>

-----

Assuming the image ordering for Oracle Linux 7.4 is here to stay, users would be better off passing the image name in directly in the data source like so:

```
data "baremetal_core_images" "ImageOCID" {
    compartment_id = "${var.compartment_ocid}"
    display_name = "Oracle-Linux-7.4-2017.08.25-0"
}
```

briangustafson:
Thanks for reporting. As mentioned in #342, we're looking into the right long term fix for specifying images. For now, I believe that your suggestion of switching the examples to filter on display_name is a good one.

tomzepf-oracle:
This has just broken quite a bit of our terraform, and a number of examples in the docs/examples directory.

briangustafson:
I've merged a change to update the examples to filter on display_name, as suggested by @jlamillan. Each example also includes the following note: "This technique is for example purposes only. The results of oci_core_images may change over time for Oracle-provided images, so the only sure way to get the correct OCID is to supply it directly."

We're looking into the correct long term solution to this problem.

briangustafson:
A long-term solution to this issue is being tracked with #352.

PR #346, opened 2017-11-01
Filter image on display name

briangustafson:
This is temporary fix for issue #345. As noted in each example usage
of oci_core_images, the images returned may change over time - a longer
term solution is in the works.

I've applied a sampling of these examples, but I have run plan on all of them which should find any mistakes with variable naming.

### DIFF
--- a/data_source_obmcs_core_instances_test.go
+++ b/data_source_obmcs_core_instances_test.go
@@ -52,9 +52,7 @@ func (s *DatasourceCoreInstanceTestSuite) SetupTest() {
 
 	data "oci_core_images" "t" {
 		compartment_id = "${var.compartment_id}"
-		operating_system = "Oracle Linux"
-		operating_system_version = "7.4"
-		limit = 1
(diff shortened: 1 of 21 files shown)


data_source_obmcs_core_instances_test.go
```diff
@@ -52,9 +52,7 @@ func (s *DatasourceCoreInstanceTestSuite) SetupTest() {
 
 	data "oci_core_images" "t" {
 		compartment_id = "${var.compartment_id}"
-		operating_system = "Oracle Linux"
-		operating_system_version = "7.4"
-		limit = 1
+		display_name = "Oracle-Linux-7.4-2017.10.25-0"
```
rohanbhattacharjee: can something like `display_name = "${var.InstanceImageDisplayName}"` not be used here (like you've done in the other cases)

briangustafson: I was just following the existing pattern in each test/example, where some use a variable and some hardcode.

PR #347, opened 2017-11-02
Add support for block volume fast clones

codycushing:
* Support creating clones from a running instance or a volume backup

Tests run: 
`make test run=Test.*CoreVolume.*`

specifically: 
* TestResourceCoreVolumeTestSuite
* TestResourceCoreVolumeBackupTestSuite
* TestDatasourceCoreVolumeTestSuite

### DIFF
--- a/docs/datasources/core/volumes.md
+++ b/docs/datasources/core/volumes.md
@@ -38,3 +38,4 @@ The following attributes are exported:
 * `state` - The current state of the volume. Allowed values are: [PROVISIONING, RESTORING, AVAILABLE, TERMINATING, TERMINATED, FAULTY]
 * `size_in_gbs` - The size of the volume, in GBs. The size must be a multiple of 1024.
 * `time_created` - The date and time the Volume was created, in the format defined by RFC3339.  Example: `2016-08-25T21:10:29.600Z`.
+* `source_details` - Specifies the volume source details for a new Block Volume.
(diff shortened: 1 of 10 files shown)


docs/datasources/core/volumes.md
```diff
@@ -38,3 +38,4 @@ The following attributes are exported:
 * `state` - The current state of the volume. Allowed values are: [PROVISIONING, RESTORING, AVAILABLE, TERMINATING, TERMINATED, FAULTY]
 * `size_in_gbs` - The size of the volume, in GBs. The size must be a multiple of 1024.
 * `time_created` - The date and time the Volume was created, in the format defined by RFC3339.  Example: `2016-08-25T21:10:29.600Z`.
+* `source_details` - Specifies the volume source details for a new Block Volume.
```
briangustafson: The API docs are somewhat more descriptive: "The volume source, either an existing volume in the same Availability Domain or a volume backup. If null, an empty volume is created." 
Should also point to [Source Details](https://docs.us-phoenix-1.oraclecloud.com/api/#/en/iaas/20160918/requests/CreateVolumeDetails) as you do in the resource.

codycushing: The resources/core/volume.md links to the source details, here that info is less relevant since it is merely a read only property that may or may not be defined.

PR #348, opened 2017-11-02
PoC examples directory refactor

codycushing:
* Refactors Identity examples into single plan
  * Less dependency setup per resource
* Adds relevant datasource example for each resource


docs/examples/identity/group.tf
```diff
@@ -0,0 +1,26 @@
+/*
+ * This example file shows how to create a group and add a user to it. 
+ */
+
+resource "oci_identity_group" "group1" {
+  name = "group1"
```
briangustafson: -tf-group

docs/examples/identity/user.tf
```diff
@@ -0,0 +1,49 @@
+/*
+ * This example file shows how to create a user, add an api key, and output their password. 
+ */
+
+resource "oci_identity_user" "user1" {
+  name = "user1"
```
briangustafson: -tf-user

docs/examples/identity/readme.md
```diff
@@ -0,0 +1,15 @@
+    #     ___  ____     _    ____ _     _____
+    #    / _ \|  _ \   / \  / ___| |   | ____|
+    #   | | | | |_) | / _ \| |   | |   |  _|
+    #   | |_| |  _ < / ___ | |___| |___| |___
+    #    \___/|_| \_/_/   \_\____|_____|_____|
+***
+## Identity Resource and Datasource Examples
+This example demonstrates the following Identity concepts:
+* Define multiple OCI providers targeting different regions
```
briangustafson: I think that we should have a separate place for examples of general concepts like this. Mixing them into the service/resource examples will make them less discoverable. I would image something like:
- docs/examples
  - identity
  - compute
  - other services
  - concepts
    - multi_region
    - filtering

docs/examples/identity/group.tf
codycushing: I was thinking about getting away from the -tf- prefix in examples since it might be unnecessarily inscrutable for casual example browsers.

docs/examples/identity/readme.md
codycushing: I think those concept examples should certainly be represented fully and distinctly, I don't have an issue including something like filtering in these examples though if it makes the files a little more meaningful without adding unnecessary complexity. So multi-region is probably a good removal, though it results in a very basic AD file.

docs/examples/identity/group.tf
briangustafson: Yeah, I agree with that. Still, group1 isn't a great name. Ideally it should be clear to someone who finds this in the console later on where it came from. Maybe TerraformExampleGroup

docs/examples/identity/readme.md
briangustafson: agree (a basic AD file seems fine)

codycushing:
Moving to internal repo for final work and review

PR #351, opened 2017-11-02
SQL Always On Availability Groups solution example

shilmover:
This is in preparation for the release of the SQL Always On white papers built on top of this TF


shilmover:
SQL Always On PR

shilmover:
Working on the Readme currently.  Also, This should not go live until the document is live.  I will update with ETA for doc.

docs/solutions/sql_server_always_on_availability_groups/configuration.tf
```diff
@@ -0,0 +1,106 @@
...
+variable "admin_shape" {
+  default = "VM.Standard1.4"
+}
+
+variable "witness_shape" {
+  default = "VM.Standard1.4"
+}
+
+locals {
+  // Subnet prefix declaration  // If not set otherwise, subnets will be auto generated.  // The first /21 block will be assigned to the DMZ-prefix.  // Second block to the Active Directory.  // Third for the SQL servers. Defaults are declared in the include.tf file.
```
briangustafson: strange formatting here

docs/solutions/sql_server_always_on_availability_groups/modules/iam/compartment/main.tf
```diff
@@ -0,0 +1,6 @@
+resource "oci_identity_compartment" "compartment" {
+  count       = "${var.compartment_id == "" ? 1 : 0}"
```
briangustafson: This is a pretty neat work around, but I don't believe it's necessary. The compartment resource will actually look for a compartment with the specified name, and if it already exists it will just use that one instead of creating a new one. So I would specify only name and description here, and once you do that there's probably not a good reason to make this a module either.

docs/solutions/sql_server_always_on_availability_groups/modules/instances/Sql/main.tf
```diff
@@ -0,0 +1,54 @@
...
+    subnet_id        = "${var.subnets["${count.index}"]}"
+    assign_public_ip = false
+  }
+
+  timeouts {
+    create = "60m"
+  }
+
+  provisioner "local-exec" {
+    command = "sleep 10"
```
briangustafson: What's the reason for this?

docs/solutions/sql_server_always_on_availability_groups/modules/instances/Sql/main.tf
```diff
@@ -0,0 +1,54 @@
+resource "oci_core_instance" "instance" {
+  count               = "${var.ad_count}"
+  availability_domain = "${lookup(data.oci_identity_availability_domains.ADs.availability_domains["${count.index}"],"name")}"
+  compartment_id      = "${var.compartment_id}"
+  display_name        = "${var.dns_label}${"${count.index}" + 1}"
+
+  #hostname_label      = "${var.dns_label}${"${count.index}" + 1}"
```
briangustafson: remove

docs/solutions/sql_server_always_on_availability_groups/terraform.tfvars
```diff
@@ -0,0 +1,18 @@
+tenancy_id = "<tenancy OCID>"
```
briangustafson: All of our other examples are using "tenancy_ocid". It's useful to be consistent so that we can users can export a common set of variables for any example.

docs/solutions/sql_server_always_on_availability_groups/terraform.tfvars
```diff
@@ -0,0 +1,18 @@
...
+
+user_id = "<user OCID>"
```
briangustafson: All of our other examples are using "user_ocid".

docs/solutions/sql_server_always_on_availability_groups/modules/network/subnet/main.tf
```diff
@@ -0,0 +1,32 @@
+resource "oci_core_subnet" "subnet" {
+  count      = "${var.ad_count}"
+  cidr_block = "${var.cidr_block["${count.index}"]}"
+
+  // Some terraform trickery required here.
+  // Due to the fact that we need to specify placement for the Witness server, we're using additional variable ${ad_placement}
```
briangustafson: I think you mean ad_deployment?

docs/solutions/sql_server_always_on_availability_groups/configuration.tf
```diff
@@ -0,0 +1,106 @@
+// vcn_dns_name is the outermost prefix in the <vcn_dns_name>.oraclevcn.com domain.
```
briangustafson: I can't seem to comment on ConfigurationFile.ini since github thinks it's binary, but where is that file used? I'm not seeing it.

briangustafson: Ah, right, I'm forgetting that these are all Windows instances, and we don't currently have a way of automating configuration in that case... so presumably your readme/doc will explain how to log in and set things up.

docs/solutions/sql_server_always_on_availability_groups/configuration.tf
shilmover: Strange in what way?

docs/solutions/sql_server_always_on_availability_groups/modules/iam/compartment/main.tf
shilmover: It actually is necessary due to to a bug in Terraform. Not putting this causes it to fail :)

docs/solutions/sql_server_always_on_availability_groups/modules/instances/Sql/main.tf
shilmover: Remove the hostname_label?

docs/solutions/sql_server_always_on_availability_groups/modules/instances/Sql/main.tf
shilmover: We found that TF would fail at times due to the order in which things are built. This solves this

docs/solutions/sql_server_always_on_availability_groups/configuration.tf
shilmover: That is correct. The .ini file is used to setup SQL on the windows instances and is not used by TF

docs/solutions/sql_server_always_on_availability_groups/modules/network/subnet/main.tf
shilmover: yes. the Witness instance can exist in any AD

docs/solutions/sql_server_always_on_availability_groups/terraform.tfvars
shilmover: Sounds good. I will change

docs/solutions/sql_server_always_on_availability_groups/terraform.tfvars
shilmover: Ditto

docs/solutions/sql_server_always_on_availability_groups/configuration.tf
briangustafson: Shouldn't each // be the start of a new line?

docs/solutions/sql_server_always_on_availability_groups/modules/instances/Sql/main.tf
briangustafson: It's commented out... is this here in case users want to add it in? If so, just put a comment above it saying so.

docs/solutions/sql_server_always_on_availability_groups/modules/instances/Sql/main.tf
briangustafson: I've opened an internal bug to track this - this should not be required.

docs/solutions/sql_server_always_on_availability_groups/configuration.tf
shilmover: Ah... Fixed...

docs/solutions/sql_server_always_on_availability_groups/modules/iam/compartment/main.tf
briangustafson: I haven't tried with this config, but testing a similar approach in another config gives me the error "value of 'count' cannot be computed", due to issue https://github.com/hashicorp/terraform/issues/10857. 

If there resource already exists I don't see the error, but if I destroy and apply again, then I see it. In this case you would want to test where compartment ID is and is not set on a fresh apply.

Gestas:
@briangustafson We're picking this back up, I want to ship it the end of February. Do you have any blocking issues/questions?

Gestas:
@briangustafson we're picking this back up, I want to ship the end of February. Do you have any blocking concerns/questions?

docs/solutions/sql_server_always_on_availability_groups/configuration.tf
```diff
@@ -0,0 +1,112 @@
...
+  default = "1.1.1.1"
+}
+
+// In which AD deploy the Witness servers
+variable "witness_deployment" {
+  default = "2"
+}
+
+// Windows 2012
+variable "image_id" {
```
briangustafson: This will change with the region - see the recommended pattern at https://github.com/oracle/terraform-provider-oci/blob/master/docs/examples/compute/multi_vnic/multi_vnic.tf#L22

docs/solutions/sql_server_always_on_availability_groups/include.tf
```diff
@@ -0,0 +1,55 @@
...
+variable "compartment_id" {}
+variable "user_id" {}
+variable "fingerprint" {}
+variable "private_key_path" {}
+variable "private_key_password" {}
+variable "region" {}
+
+variable "home_region" {}
+
+//True by default. Solve inconsistency problems but creates performance issue on destroy.
```
briangustafson: The perf issue on destroy has been fixed for a while - let's get rid of this variable.

docs/solutions/sql_server_always_on_availability_groups/modules/iam/compartment/main.tf
briangustafson: This comment still needs to be addressed. I would be in favor or removing count and var.compartment_id. Actually, I would completely get rid of the compartment module - just use the resource directly.

docs/solutions/sql_server_always_on_availability_groups/modules/network/securitylist/main.tf
```diff
@@ -0,0 +1,430 @@
...
+    {
+      protocol    = "${var.proto_number_All}"
+      destination = "0.0.0.0/0"
+    },
+  ]
+
+  ingress_security_rules = [
+    {
+      tcp_options {
+        "max" = 3389
```
briangustafson: Would be good to have some comments in this file explaining where these rules came from.

docs/solutions/sql_server_always_on_availability_groups/outputs.tf
```diff
@@ -0,0 +1,53 @@
+// Outputs declaration
+
+#output "sql_ids" {
```
briangustafson: Why are these commented out? If not needed, remove.

shilmover:
Closing

PR #323, opened 2017-10-20
KVM Sample code

lucassrg:
Terraform Sample code for KVM baremetal and KVM nested

PR #281, opened 2017-10-03
Bring your own hypervisor sample code

lucassrg:
This sample uses Terraform to demonstrate the Bring Your Own Hypervisor model to deploy custom KVM based Virtual Machines on top of a KVM hypervisor on Oracle Cloud Infrastructure.

PR #353, opened 2017-11-07
kvm hypervisor code

lucassrg:
Initial version of KVM hypervisor solution for both nested kvm and bare metal.

Also moved kvm sample code from examples to solutions.


docs/solutions/hypervisors/kvm/kvm-baremetal/README.md
```diff
@@ -0,0 +1,114 @@
...
+------------
+
+- Access to OCI environment
+- [Generated OCI keys and OCIDs](https://docs.us-phoenix-1.oraclecloud.com/Content/API/Concepts/apisigningkey.htm)
+- Terraform 0.10.x+
+
+
+### Usage
+
+- Update `terraform.tfvars` with the required information. Terraform automatically loads them to populate variables, but you can also use the -var-file flag directly to specify a file. These files are the same syntax as Terraform configuration files. And like Terraform configuration files, these files can also be JSON.
```
briangustafson: terraform.tfvars is not in the PR. Most of our complex examples include an env-vars - I recommend doing that.

docs/solutions/hypervisors/kvm/kvm-baremetal/variables.tf
```diff
@@ -0,0 +1,28 @@
...
+variable "user_ocid" {}
+variable "fingerprint" {}
+variable "compartment_ocid" {}
+variable "region" {}
+variable "ssh_public_key_path" {}
+variable "ssh_private_key_path" {}
+
+#Template used variables
+variable "prefix" {}
+variable "availability_domain" {}
```
briangustafson: Would be useful to set defaults for all variables where a default value can reasonably be used.

docs/solutions/hypervisors/kvm/kvm-baremetal/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+We don't recommend saving usernames and passwords to version control, but you can create a local secret variables file and use -var-file to load it.
+
+Most examples use the same set of environment variables so you only need to do this once. There's a sample file available on this sample code for your reference in addition to the snippet below:
+
+```
+###############################
+#### Environment variables ####
```
briangustafson: Missing a few variables.... var.kvm_host_subnet_cidr_block, var.prefix, var.vcn_cidr_block. (If you add defaults and an env-vars file though, I don't think there's any need to list variables in the README.)

docs/solutions/hypervisors/kvm/kvm-baremetal/datasource.tf
```diff
@@ -0,0 +1,27 @@
+data "oci_identity_availability_domains" "ads" {
+  compartment_id = "${var.tenancy_ocid}"
+}
+
+data "oci_core_images" "base-image" {
+  compartment_id = "${var.compartment_ocid}"
+  display_name   = "Oracle-Linux-7.4-2017.10.25-0"
+}
+
+data "oci_core_shape" "supported_shapes" {
```
briangustafson: Is the goal here to ensure that users are using BM shapes? This doesn't seem necessary to me - I would just have them specify the shape directly and document that it must be BM (as is already done).

docs/solutions/hypervisors/kvm/kvm-baremetal/modules/kvm-installer-baremetal/README.md
```diff
@@ -0,0 +1,128 @@
+KVM Installation module
```
briangustafson: The baremetal and nested examples are very similar. Not blocking, but might make sense to have a README one level up that has some of the common info and explains the difference.

briangustafson:
Just for reference, linking to previous iterations at  #323 and #281.

docs/solutions/hypervisors/kvm/kvm-baremetal/README.md
lucassrg: For some reason the file disappeared on my local workspace, I will push it. However, I would say env-vars is considered an anti-pattern on Terraform. First because it's global in your shell and in case you have multiple terraform projects you have likely a great chance of running the wrong env-vars set, and having a .tfvars file which is local to your project you mitigate that risk. In addition to that, there are some security considerations, since setting env vars in a shell could potentially allow access from other applications to unencrypted keys saved in the shell. Ideally, we should set those keys at the command line when running terraform.

docs/solutions/hypervisors/kvm/kvm-baremetal/datasource.tf
lucassrg: Yes, this is a safe-guard - just  to ensure users are using the appropriate shape, otherwise launching the KVM domain fails without any hint. That happened to me when I was modifying the code and set the wrong shape value for the nested kvm.

docs/solutions/hypervisors/kvm/kvm-baremetal/modules/kvm-installer-baremetal/README.md
lucassrg: Agreed. I just created another README as you suggested.

docs/solutions/hypervisors/kvm/README.md
```diff
@@ -0,0 +1,20 @@
+KVM Virtualization
+===============================
+This folder contains sample terraform templates to allow you to run a KVM hypervisor on Oracle Cloud Infrastructure, based on the [Installing and Configuring KVM on Bare Metal Instances with Multi-VNIC] (https://docs.us-phoenix-1.oraclecloud.com/Content/Resources/Assets/installing_kvm_multi_vnics.pdf) white-paper and [A Simple Guide to Nested KVM Virtualization on Oracle Cloud Infrastructure](https://blogs.oracle.com/cloud-infrastructure/nested-kvm-virtualization-on-oracle-iaas) blog post.
```
briangustafson: The link for [Installing and Configuring KVM on Bare Metal Instances with Multi-VNIC] isn't showing up correctly - probably because of the space between ] and (.

docs/solutions/hypervisors/kvm/kvm-nested/terraform.tfvars
```diff
@@ -0,0 +1,48 @@
+###############################
+#### Environment variables ####
+###############################
+
+tenancy_ocid="ocid1.tenancy.oc1..abcabcababccbcbaabcabababcbabababcbababcbabababacbcabcabcabac4uq"
```
briangustafson: Please switch these to placeholder text, similar to what is done in env-vars files for other examples such as https://github.com/oracle/terraform-provider-oci/blob/master/docs/solutions/rhel74_image/env-vars. 

Generally, if there's a sensible default that the user could realistically use, then set it to that. If the user must set a value of their own (as with tenancy_ocid, user, fingerprint, etc), then set placeholder text.

docs/solutions/hypervisors/kvm/kvm-baremetal/terraform.tfvars
```diff
@@ -0,0 +1,47 @@
+###############################
+#### Environment variables ####
+###############################
+tenancy_ocid="ocid1.tenancy.oc1..abcabcababccbcbaabcabababcbabababcbababcbabababacbcabcabcabac4uq"
```
briangustafson: Same comment about placeholder text as with other terraform.tfvars file.

docs/solutions/hypervisors/kvm/kvm-nested/terraform.tfvars
lucassrg: @briangustafson I fixed all the code according to all your recommendations. I replaced the tfvar files by env-vars. Now It's look like OK. Thanks for your time reviewing the code.

docs/solutions/hypervisors/kvm/kvm-baremetal/main.tf
```diff
@@ -0,0 +1,74 @@
+provider "oci" {
+  tenancy_ocid     = "${var.tenancy_ocid}"
+  user_ocid        = "${var.user_ocid}"
+  fingerprint      = "${var.fingerprint}"
+  private_key_path = "${var.ssh_private_key_path}"
```
briangustafson: The ssh key should be different from the key used to make requests. (See the "important" message at https://docs.us-phoenix-1.oraclecloud.com/Content/API/Concepts/apisigningkey.htm.) Should have a separate variable for private_key_path.

docs/solutions/hypervisors/kvm/kvm-nested/main.tf
```diff
@@ -0,0 +1,61 @@
+provider "oci" {
+  tenancy_ocid     = "${var.tenancy_ocid}"
+  user_ocid        = "${var.user_ocid}"
+  fingerprint      = "${var.fingerprint}"
+  private_key_path = "${var.ssh_private_key_path}"
```
briangustafson: ditto

docs/solutions/hypervisors/kvm/kvm-baremetal/main.tf
lucassrg: This private key is used on the template to perform `remote-exec` operations (SSH/SCP) and used within a [connection](https://www.terraform.io/docs/provisioners/connection.html) block.  So, per my understanding, this should be the RSA key, not the PEM format used by the SDK. I just need remove it from the provider declaration and perhaps provider a better name to differentiate from the `private_key_path` variable.

docs/solutions/hypervisors/kvm/kvm-nested/main.tf
briangustafson: Correct - see for example https://github.com/oracle/terraform-provider-oci/blob/master/docs/solutions/omc_monitored_server/env-vars

docs/solutions/hypervisors/kvm/kvm-baremetal/main.tf
briangustafson: Correct - see for example https://github.com/oracle/terraform-provider-oci/blob/master/docs/solutions/omc_monitored_server/env-vars

briangustafson:
@lucassrg I believe this is just waiting on some minor updates on your side, per my previous comments, and then this ought to be good to go.

lucassrg:
Hi @briangustafson, that's correct. Actually I just need to finish some tests to validate my changes. I'll find some time to finish that today.

lucassrg:
@briangustafson, all the suggestions were implemented, including ssh key variables, selected compute shape and env-vars script.

docs/solutions/hypervisors/kvm/kvm-baremetal/main.tf
```diff
@@ -2,7 +2,7 @@ provider "oci" {
   tenancy_ocid     = "${var.tenancy_ocid}"
   user_ocid        = "${var.user_ocid}"
   fingerprint      = "${var.fingerprint}"
-  private_key_path = "${var.ssh_private_key_path}"
+  private_key_path = "${var.ssh_api_private_key_path}"
```
briangustafson: This isn't really an "ssh" key... in all other solutions and examples this is called private_key_path and also has a private_key_password, such that the provider looks like:

provider "oci" {
  tenancy_ocid = "${var.tenancy_ocid}"
  user_ocid = "${var.user_ocid}"
  fingerprint = "${var.fingerprint}"
  private_key_path = "${var.private_key_path}"
  region = "${var.region}"
  private_key_password = "${var.private_key_password}"
}

docs/solutions/hypervisors/kvm/kvm-nested/main.tf
```diff
@@ -2,7 +2,7 @@ provider "oci" {
   tenancy_ocid     = "${var.tenancy_ocid}"
   user_ocid        = "${var.user_ocid}"
   fingerprint      = "${var.fingerprint}"
-  private_key_path = "${var.ssh_private_key_path}"
+  private_key_path = "${var.ssh_api_private_key_path}"
```
briangustafson: Same comment about naming

docs/solutions/hypervisors/kvm/kvm-baremetal/main.tf
```diff
@@ -0,0 +1,73 @@
...
+  region           = "${var.region}"
+}
+
+resource "oci_core_instance" "kvm-host-instance" {
+  availability_domain = "${lookup(data.oci_identity_availability_domains.ads.availability_domains[var.availability_domain - 1],"name")}"
+  compartment_id      = "${var.compartment_ocid}"
+  display_name        = "${var.prefix}-kvm-host"
+  image               = "${lookup(data.oci_core_images.base-image.images[0], "id")}"
+  shape               = "${instance_bm_shape}"
```
briangustafson: missing "var."

docs/solutions/hypervisors/kvm/kvm-baremetal/main.tf
```diff
@@ -0,0 +1,73 @@
+provider "oci" {
+  tenancy_ocid     = "${var.tenancy_ocid}"
+  user_ocid        = "${var.user_ocid}"
+  fingerprint      = "${var.fingerprint}"
+  private_key_path = "${var.ssh_api_private_key_path}"
```
briangustafson: This isn't really an "ssh" key... in all other solutions and examples this is called private_key_path and also has a private_key_password, such that the provider looks like:

provider "oci" {
tenancy_ocid = "${var.tenancy_ocid}"
user_ocid = "${var.user_ocid}"
fingerprint = "${var.fingerprint}"
private_key_path = "${var.private_key_path}"
region = "${var.region}"
private_key_password = "${var.private_key_password}"
}

docs/solutions/hypervisors/kvm/kvm-nested/main.tf
```diff
@@ -0,0 +1,60 @@
+provider "oci" {
+  tenancy_ocid     = "${var.tenancy_ocid}"
+  user_ocid        = "${var.user_ocid}"
+  fingerprint      = "${var.fingerprint}"
+  private_key_path = "${var.ssh_api_private_key_path}"
```
briangustafson: same comment about naming

docs/solutions/hypervisors/kvm/kvm-nested/main.tf
```diff
@@ -0,0 +1,60 @@
...
+  region           = "${var.region}"
+}
+
+resource "oci_core_instance" "kvm-host-instance" {
+  availability_domain = "${lookup(data.oci_identity_availability_domains.ads.availability_domains[var.availability_domain - 1],"name")}"
+  compartment_id      = "${var.compartment_ocid}"
+  display_name        = "${var.prefix}-kvm-host"
+  image               = "${lookup(data.oci_core_images.base-image.images[0], "id")}"
+  shape               = "${var.instance_vm_shape)}"
```
briangustafson: remove )

lucassrg:
For some reason my code broke after updating Terraform to the latest version.

lucassrg:
I fixed that issue related to the image by specifying the image OCID direct to the oci_instance resource. Now I have another issue with one of the scripts I created to set up the KVM.

docs/solutions/hypervisors/kvm/kvm-baremetal/README.md
briangustafson: I'm still not seeing the terraform.tfvars files in the latest review... I'm fine with including that, but if it's not included (and env-vars only is given) then this line needs to be removed.

Ah, wait, so the content later in the readme is the template for terraform.tfvars? This needs to be a bit clearer.

docs/solutions/hypervisors/kvm/kvm-baremetal/variables.tf
```diff
@@ -0,0 +1,90 @@
...
+}
+
+#KVM Host instance shape. Only BM Shapes are supported on this example
+variable "instance_bm_shape" {
+  default = "BM.Standard1.36"
+}
+
+#URL of your image file for downloading (you can place your image in the object storage!)
+variable "kvm_image_url" {
+  default = "https://<my-qcow2-image-url>"
```
briangustafson: I would think that this shouldn't have a default, since it won't work unless you provide a different value.

docs/solutions/hypervisors/kvm/kvm-nested/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+#Customer Identifier to be used before the resources name
+customer_name = "mycustomer"
+
+#availability_domain number - For AD1 uses 1. For AD2, uses 2, For AD3, uses 3
+availability_domain = "1"
+
+#Oracle Linux 7.4 Image OCID. Image OCI for specific region is available in the documentation:
+# https://docs.us-phoenix-1.oraclecloud.com/Content/Compute/References/images.htm
+variable "image_ocid" {
```
briangustafson: Recommend to make this a map with region as the key:

variable "image_ocid" {
  type = "map"
  default = {
    // Oracle-provided image "Oracle-Linux-7.4-2017.12.18-0"
    // See https://docs.us-phoenix-1.oraclecloud.com/Content/Resources/Assets/OracleProvidedImageOCIDs.pdf
    us-phoenix-1 = "ocid1.image.oc1.phx.aaaaaaaasc56hnpnx7swoyd2fw5gyvbn3kcdmqc2guiiuvnztl2erth62xnq"
    us-ashburn-1 = "ocid1.image.oc1.iad.aaaaaaaaxrqeombwty6jyqgk3fraczdd63bv66xgfsqka4ktr7c57awr3p5a"
    eu-frankfurt-1 = "ocid1.image.oc1.eu-frankfurt-1.aaaaaaaayxmzu6n5hsntq4wlffpb4h6qh6z3uskpbm5v3v4egqlqvwicfbyq"
  }
}

Issue #354, opened 2017-11-07
Update readme to specify TF_VAR_compartment_id and other necessary vars for testing


alexng-canuck:
Some test suites such as TestDatasourceCoreConsoleHistoryTestSuite, require a compartment_id environment variable to be configured. This step is missing from the README.

Without this step, we would get a 404 error that's not helpful for root causing the issue (unless you dove into the debug logs). Let's add those required env vars in the README.

PR #358, opened 2017-11-08
Add compartment_ocid environment var support in tests

alexng-canuck:
Tests are using the compartment_id env var while examples use the
compartment_ocid environment var. Avoid having to specify two variables
by using compartment_ocid in tests as well. Tests will still support the
compartment_id name to avoid breaking folks who have this setting.

This should simplify the README steps, which have also been updated to
include the compartment_ocid env var.

This addresses #354

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,8 +88,9 @@ In your ~/.bash_profile set these variables
 ```
 export TF_VAR_tenancy_ocid=
 export TF_VAR_user_ocid=
+export TF_VAR_compartment_ocid=<The tenancy OCID can be used as the compartment OCID of your root compartment>
 export TF_VAR_fingerprint=
-export TF_VAR_private_key_path=<fully qualified path>`
+export TF_VAR_private_key_path=<fully qualified path>
 ```
 
 Once you've set these values open a new terminal or source your profile changes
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -101,6 +102,7 @@ $ source ~/.bash_profile
 ```
 setx TF_VAR_tenancy_ocid <value>
 setx TF_VAR_user_ocid <value>
+setx TF_VAR_compartment_ocid <value>
```
briangustafson: We also mention setting the compartment on line 114. That line could be removed with this change... although actually I see that the compute/instance example we're telling users to run has other variables that must be set for it to work. So that's really not the right example to point them to.

Really, I think a bigger cleanup is needed here. For one, the other variables listed here are all ones that are required for the provider - that's not true of compartment_ocid. However, these variable names are also just a convention that we use in our examples - there's nothing magic about them. We do though, have magic variable names that just work with the provider, see https://github.com/oracle/terraform-provider-oci/blob/master/provider.go#L53, and we're not documenting that anywhere.

briangustafson: Until we do a bigger change around this, I think I'll recommend keeping what you have, switching the example we reference to networking/vcn (it's very simple example that only needs the variables mentioned here, I think, should verify) and remove line 114.

alexng-canuck: I'll take your suggestions. From a test perspective, the compartment_ocid is a defacto required variable; which is why I decided to treat it as such. Perhaps as part of the bigger clean-up we may need to add setup steps that are relevant to running the tests.

alexng-canuck: @briangustafson I've addressed your feedback in the latest commit

PR #360, opened 2017-11-13
Add glusterfs sample configuration

alexng-canuck:
This adds a sample configuration for creating a glusterfs volume that's
replicated across 3 ADs. This sample is taken from gilson.melo@oracle.com
with the following changes:

- Add networking resources required by the glusterfs servers and clients
- Rename/Introduce some variables to parameterize the sample more consistently
with the other samples
- Added a README file to explain the configuration

### DIFF
--- a/docs/solutions/glusterfs/README.md
+++ b/docs/solutions/glusterfs/README.md
@@ -1,0 +1,45 @@
+    #     ___  ____     _    ____ _     _____
+    #    / _ \|  _ \   / \  / ___| |   | ____|
+    #   | | | | |_) | / _ \| |   | |   |  _|
+    #   | |_| |  _ < / ___ | |___| |___| |___
+    #    \___/|_| \_/_/   \_\____|_____|_____|
+***
+## GlusterFS Infrastructure
+This configuration creates a glusterfs volume that is replicated across 3 GlusterFS servers.
+
+It creates a VCN with a route table, internet gateway, and security list.
+The VCN spans 3 ADs with each AD containing a subnet and 2 instances: a GlusterFS server and a GlusterFS client. 
+
+### Using this example
+* Update env-vars with the required information. Most examples use the same set of environment variables so you only need to do this once.
+* Source env-vars
+  * `$ . env-vars`
+* Update `variables.tf` with your instance options.
+* Update `./userdata/bootstrap` by replacing instances of `baremetal.oraclevcn.com` with `[Your VCN's DnsLabel].oraclevcn.com`.
+
+### Files in the configuration
+
+#### `env-vars`
+Is used to export the environmental variables used in the configuration. These are usually authentication related, be sure to exclude this file from your version control system. It's typical to keep this file outside of the configuration.
+
+Before you plan, apply, or destroy the configuration source the file -  
+`$ . env-vars`
+
+#### `compute.tf`
+Defines the compute resources
+
+#### `networking.tf`
+Defines the virtual cloud network resources used in the configuration
+
+#### `variables.tf`
+Defines the variables used in the configuration
+
+#### `datasources.tf`
+Defines the datasources used in the configuration
+
+#### `provider.tf`
+Specifies and passes authentication details to the OCI TF provider
+
+#### `./userdata/bootstrap`
+The script gets injected into an instance on launch.
+The script configures the glusterfs volumes on each server and sets up the glusterfs clients.
--- a/docs/solutions/glusterfs/compute.tf
+++ b/docs/solutions/glusterfs/compute.tf
@@ -1,0 +1,85 @@
+resource "oci_core_instance" "GlusterServerInstance" {
+  availability_domain = "${oci_core_subnet.SubnetAD1.availability_domain}"
+  compartment_id = "${var.compartment_ocid}"
+  display_name = "glusterfs-server1"
+  hostname_label = "glusterfs-server1"
+  image = "${lookup(data.oci_core_images.ServerImageList.images[0], "id")}"
+  shape = "${var.ServerInstanceShape}"
+  subnet_id = "${oci_core_subnet.SubnetAD1.id}"
+  metadata {
+    ssh_authorized_keys = "${var.ssh_public_key}"
+    user_data = "${base64encode(file(var.ServerBootStrapFile))}"
+  }
+}
+
+resource "oci_core_instance" "GlusterServerInstance2" {
+  availability_domain = "${oci_core_subnet.SubnetAD2.availability_domain}"
+  compartment_id = "${var.compartment_ocid}"
+  display_name = "glusterfs-server2"
+  hostname_label = "glusterfs-server2"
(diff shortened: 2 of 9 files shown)


alexng-canuck:
Adding @gilsonmelo, who created the basis of this sample to help review.

docs/examples/clusters/glusterfs/userdata/bootstrap
```diff
@@ -0,0 +1,152 @@
+#!/bin/bash
```
kohashim: General comments:
1. Indentation looks off so not easy to read.
1. I would break into at least two files such as bootstrap_server and bootstrap_client.
1. I would try to avoid duplicated code by parameterizing the common logic by introducing local vars which are set based on the similar case statement.
1. Just curious, the device and dir names "brick[123]" need to be numbered?  Can't it be the same name among the three servers?

docs/examples/clusters/glusterfs/README.md
```diff
@@ -0,0 +1,45 @@
+    #     ___  ____     _    ____ _     _____
+    #    / _ \|  _ \   / \  / ___| |   | ____|
+    #   | | | | |_) | / _ \| |   | |   |  _|
+    #   | |_| |  _ < / ___ | |___| |___| |___
+    #    \___/|_| \_/_/   \_\____|_____|_____|
+***
+## GlusterFS Infrastructure
+This configuration creates a glusterfs volume that is replicated across 3 GlusterFS servers.
+
+It creates a VNC with a route table, internet gateway, and security list.
```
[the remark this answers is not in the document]
alexng-canuck: Good catch on the VNC typos.

docs/examples/clusters/glusterfs/env-vars
```diff
@@ -0,0 +1,13 @@
+### Authentication details
+export TF_VAR_tenancy_ocid="<tenancy OCID>"
+export TF_VAR_user_ocid="<user OCID>"
+export TF_VAR_fingerprint="<PEM key fingerprint>"
+export TF_VAR_private_key_path="<path to the private key that matches the fingerprint above>"
+export TF_VAR_private_key_password="<password for your private key>"
+export TF_VAR_private_key=
```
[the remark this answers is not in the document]
alexng-canuck: You are right, this variable isn't used. We should remove this.

docs/examples/clusters/glusterfs/networking.tf
```diff
@@ -0,0 +1,77 @@
+variable "VPC-CIDR" {
```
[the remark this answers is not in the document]
alexng-canuck: Good point. We should hard code the CIDR into the VCN resource since the subnets are already hardcoded.

docs/examples/clusters/glusterfs/networking.tf
```diff
@@ -0,0 +1,77 @@
...
+  }
+}
+
+resource "oci_core_security_list" "Subnet" {
+  compartment_id = "${var.compartment_ocid}"
+  display_name = "Subnet"
+  vcn_id = "${oci_core_virtual_network.CompleteVCN.id}"
+  egress_security_rules = [{
+    protocol = "all"
+    destination = "0.0.0.0/0"
```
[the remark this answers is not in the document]
alexng-canuck: More for accessibility purposes. So I can SSH into those instances to verify and dig into how gluster works.

docs/examples/clusters/glusterfs/userdata/bootstrap
alexng-canuck: 1. Good catch, need to fix the indentation settings in my IDE :)
2. I also considered this. Sounds like it would be a good idea to separate the server/client logic.
3. This is a good point. I will fix this.
4. There's no reason to have them be numbered. Will fix this.

docs/examples/clusters/glusterfs/userdata/bootstrap
```diff
@@ -0,0 +1,152 @@
...
+### 4- disable local firewall. Feel free to update this script to open only the required ports. I have a security list rule that allows communication
+### within my tenancy (subnet Source: 172.0.0.0/16  IP Protocol: All Protocols Allows: all traffic for all ports
+### 5- install and configure a gluster volume called glustervol1 using server1-brick1, server2-brick2 and server3-brick3 LVs (replicas)
+### 6- creates a local mount point in the glusterFS clients
+### 7- add a local entry in the clients Fstab with '_netdev' attribute. IMPORTANT to avoid boot issues.
+### 
+### Feel free to contact me in case of any questions - Gilson Melo - gilson.melo@oracle.com
+######################################################################################################################################################
+exec 2>/dev/null
+glusterfshost=$(hostname -s)
```
rohanbhattacharjee: I thought Chef/Puppet are the solutions that we recommend to people for setting up a particular instance with a particular set of libraries, applications etc. Does Chef etc. not work for this glusterFs installation, or am I mistaken in my understanding of Chef etc. ?

I just saw an email from Brian which described how Docker/Kubernetes are becoming industry standards. Looks like GlusterFs can be installed using docker - https://github.com/gluster/gluster-containers, and thereby orchestrated using k8s.

I am wondering if we should, at this point, start assuming our examples to be code that promotes the state-of-the-art. I will assume that people will copy-paste this code as is. Hence if we want our customers to adopt Chef/Docker etc., our examples should do that as well.

docs/examples/clusters/glusterfs/userdata/bootstrap
```diff
@@ -0,0 +1,152 @@
...
+        yum install glusterfs glusterfs-fuse attr -y
+        mkdir /mnt/glustervol1
+        sleep 4m
+        echo "glusterfs-server3:/glustervol1 /mnt/glustervol1   glusterfs defaults,_netdev  0 0" >> /etc/fstab
+        mount -a
+        ;;
+	*)
+        echo "$glusterfshost" >>/tmp/notfound.txt
+		;;
+esac
```
rohanbhattacharjee: I would recommend not publishing this example. This code looks hard to test for correctness, it looks hard to decide if it is minimalistic (or does more stuff than what is needed), hard to decide if it's secure etc.

I would assume, if published by somebody from Oracle, that this copy will be copy-pasted.

docs/examples/clusters/glusterfs/userdata/bootstrap
kohashim: Depends on the purpose of "docs/examples".  If the directory is for Oracle to show it's recommendations, you might be right. (but I would even argue if it is true because OCI provides an infrastructure where customers can use however they want.)  On the other hand, if the purpose is just to show a simple example, I think each example should focus on only its subject (in this case glusterfs).

alexng-canuck:
Sync-up with @rohanbhattacharjee and @kohashim offline. There is a discussion to be had as a team about what standards we should have for publishing good samples. For now, those concerns are not a blocker for this particular request.

Issue #338, opened 2017-10-27
Unable to specify source ports in security list rules


darkfibre:
### Terraform Version
Terraform v0.10.7

### OCI Provider Version
terraform-provider-oci_v2.0.3

### Description:
Both ingress and egress security list rules assume the ports specified in "tcp_options" or "udp_options" are destination ports. There are situations where source ports must be specified, stateless rules in particular need to match the source port in order to permit return traffic.

You can manually fix the rule using the OCI web interface, but subsequent terraform runs will revert the change.

briangustafson:
Thanks for reporting, we'll be looking into this.

briangustafson:
This has been fixed in release 2.0.5. Please give it a try.

PR #366, opened 2017-11-20
Add source port range to security list rules

briangustafson:
Currently, TCP and UDP options can be specified with a single set of
max/min values, which correspond to the destination port range. This change
adds the ability to specify the source port range as well, issue #338.
    
Ideally, we should deprecate the existing max/min params and add a new
destination_port_range option next to the new source_port_range. Unfortunately,
a number of issue with nested schemas in Terraform are preventing support of both
of these params simultaneously. Instead, max/min params will remain, and source_port_range
will exist beside it.
    
This change also enables security lists with no rules.
    
Some updates are made to the docs, but the docs for the security list resource and data source
could really use a bigger update that is not included in this change.


docs/examples/networking/security_list/security_list.tf
```diff
@@ -20,12 +17,25 @@ provider "oci" {
   region = "${var.region}"
 }
 
+resource "oci_core_virtual_network" "vcn1" {
+  cidr_block = "10.0.0.0/16"
+  dns_label = "vnc1"
```
alexng-canuck: I assume you mean "vcn1". I see this in many of our other examples too :)

docs/resources/core/security_list.md
```diff
@@ -32,20 +32,27 @@ resource "oci_core_security_list" "t" {
...
-        stateful = true
+        stateless = true
 
         tcp_options {
-            "min" = 80
-            "max" = 82
+            source_port_range {
+                "min" = 100
+                "max" = 100
+             }
```
alexng-canuck: Minor style issue: The indentation seems off.

resource_obmcs_core_security_list.go
```diff
@@ -332,33 +321,59 @@ func (s *SecurityListResourceCrud) buildICMPOptions(conf map[string]interface{})
...
+}
+
+func buildPortRange(conf []interface{}) (portRange *baremetal.PortRange) {
+	if len(conf) > 0 && conf[0] != nil {
+		mapConf := conf[0].(map[string]interface{})
+
+		// The ok value will always be true, so we need to check against the default value instead.
+		max := mapConf["max"].(int)
+		min := mapConf["min"].(int)
+
```
alexng-canuck: Should we check that max and min are valid?
For example: min <= max and min/max are within the allowable port numbers for the protocol.

It might also be ok if the APIs give us a reasonable error message and we just bubble it up.

resource_obmcs_core_security_list.go
```diff
@@ -332,33 +321,59 @@ func (s *SecurityListResourceCrud) buildICMPOptions(conf map[string]interface{})
...
+		// Max and Min default to 0, and that is not a valid port number, so we can assume that if
+		// the value is 0 then the user has not set the port number.
+		if max != 0 || min != 0 {
```
alexng-canuck: Since both max and min are optional, what happens if someone specifies max without min or vice-versa?

Should we assume no min means everything <= max and no max means everything >= min?

alexng-canuck:
I agree with Koji here. We should figure out a way to make the source and destination port range specified in the same way. Otherwise, it would be pretty confusing without the comments.

If nesting schemas is an issue, maybe we can consider specifying it as destination_min, destination_max, source_min, source_max; and making them all optional.

briangustafson:
I'll write up the issues with supporting destination_port_range. I tried out a number of approaches that didn't quite work for various reasons that I'll detail, but it's possible that I've missed something.

resource_obmcs_core_security_list.go
briangustafson: The service does that validation as well, and gives a pretty understandable message. It would be a slightly better use experience to fail before making the service call, but at the expense of duplicating service logic and for a pretty small benefit.

What would be more beneficial would be to add a validation function in the schema definition, which could apply validations like this on a "terraform plan". I'll make that as nice to have, but not necessary for this change and not particularly high priority.

resource_obmcs_core_security_list.go
briangustafson: Specifying one without the other will fail, since the non-specified one will be 0. That's another case that would be nice to check with a validation function, until we do that the service will validate.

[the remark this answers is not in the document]
briangustafson: If you only specify one, then you will get an error from the server. However, I'd rather the plan fail than have it silently ignore the one param that the did set. (Again, we could check at the planning state with a validation function, which I view as nice to have but not necessary for now.)

rohanbhattacharjee:
I looked through the code and have some of the same comments as Koji and Alex.

Like the `min != 0 || max != 0)` statement is confusing to me ... I would have expected `min != 0 && max != 0)` ... even after reading your reply, I think it will be confusing to somebody who hasn't read this conversation.

Similarly, I would prefer the `source_port_min` etc. .. so i will wait for the doc you write.

Finally, as a general comment, I see this `interface{}` type used so often in the method signatures (e.g. `buildSourceAndDestinationPortRanges` takes in `[]interface{}`), that its hard to me to review and understand the consequences of statements like `len(conf) > 1` and `conf[0]` etc. usages are reliable and safe etc.

I see data structures like `[]map[string]interface{}`, and I don't quite like those either. At some point, I would like to sit with you and understand if I can help to improve some of these.

resource_obmcs_core_security_list.go
alexng-canuck: I guess the question is whether the writer of the tf config knows that not specifying a value means 0. A tf config writer might incorrectly assume that not specifying a value means that it's an open ended range.

Not a blocker, just some food for thought.

briangustafson:
Instead of trying to rebase this with the file refactor, I'm going to close this and open a new PR. First I'll post a straight rebase, then I'll add a commit to respond to comments.

docs/resources/core/security_list.md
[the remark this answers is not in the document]
briangustafson: fixed in new pr

resource_obmcs_core_security_list.go
briangustafson: I'll add a comment about this in the new PR. An important fact that I didn't mention is that 0 is not a valid port number, and I've verified with networking that that will never be a valid input for these params.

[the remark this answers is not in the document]
briangustafson: Will add a comment about this in the new PR.

briangustafson:
@rohanbhattacharjee - I did some very minor refactoring and commenting for the new PR. I agree that it could be clearer in places - partly security lists just have a complicated schema. For this PR I'm not going to do a significant refactor of what we have though.

PR #367, opened 2017-11-28
Fix security list resource doc

briangustafson:
This doc could use more cleanup, but this change at least gets rid of the
inaccuracies. I ran the example resource to verify that it works.

### DIFF
--- a/docs/resources/core/security_list.md
+++ b/docs/resources/core/security_list.md
@@ -32,7 +32,7 @@ resource "oci_core_security_list" "t" {
     ingress_security_rules {
         protocol = "6"
         source = "0.0.0.0/0"
-        stateful = true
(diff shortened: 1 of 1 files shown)


docs/resources/core/security_list.md
```diff
@@ -58,10 +58,10 @@ resource "oci_core_security_list" "t" {
 The following arguments are supported:
 
 * `compartment_id` - (Required) The OCID of the compartment to contain the security list.
-* `display_name` - (Required) The OCID of the VCN.
+* `display_name` - (Optional) The OCID of the VCN.
```
alexng-canuck: Fit and finish comment: We should update the description of the "display_name" argument to reflect the actual purpose of it.

PR #370, opened 2017-11-29
Source file re-organization

rohanbhattacharjee:
1. Moved most of the code under the provider directory/package
2. Made filenames shorter
3. Minor update to the Makefile so that the test are looked up in the /provider package


briangustafson:
Looks good - this is much cleaner. My only significant comment is that I believe we should leave in the service/service group prefix (“core_“, “objectstorage_“, “identity_“, etc). I think that the grouping that provides is worthwhile (and will help with name collisions between services).

(@codycushing - interested to get your feedback on this.)

AlaaShaker:
LGTM. I have a couple of questions though:
1. Is there a reason why test files aren't under a separate /test folder?
2. Is there a reason why we don't have separate /datasources & /resources folders + even shorter file names?

If this is a terraform or go convention, feel free to ignore that & I'd like to learn more about it. It just seems to me that this work is geared towards cleaning up the root folder (which makes the main page on the github repo look really long) opposed to a deep project restructure.

alexng-canuck:
I also have the same questions as @AlaaShaker.

Regarding @briangustafson comment about keeping the service group prefix, perhaps we consider creating a separate directory for each service group? Currently, it looks like we moved the existing source files en masse to the provider directory; so everything is still under a single directory. If we keep adding more resources or services in the future, we would still have the same issue of too many files under a single directory.

If my comment doesn't make sense because of some terraform or go convention, please feel free to ignore. :)

shalka:
Re-organization looks great. Instead of 'provider' maybe we follow other providers' example and use a top-level 'oci' directory? Then everything falls under 'oci' package instead of 'provider' package..? 

It's also convention to keep resource_ and data_source as pre-pended on the name (primarily for organization purposes). Ultimately, it's just a relic of file organization (and that info still appears in the filename, so I think you're good), but I agree with @briangustafson that we should keep the service name in the filename. Please do not use directories for organizing files, and just keep a flat list.

rohanbhattacharjee:
@briangustafson - added the service/service-group prefix.

@AlaaShaker @alexng-canuck - We considered various alternative organization structures. The details are covered in the JIRAs and Wiki (internally available - will ping you offline).

I had suggested we separate the tests out into its `test` directory. But we could not reach a consensus on having a separate `test` directory - so, that was not included in this PR.

Directory structures, packages and the package-import facilities in Go are somewhat cumbersome, and tricky to get right. I will forward an internal JIRA to you guys to see the discussions on it. Hence I did not go in more aggressively in organizing the code into more (coherent) directory structures. I see this code reorg as a work in progress.

@shalka - I had suggested we use the `oci` name (`aws`, `azurem` etc. are being used by other providers). But @codycushing has been suggesting the `provider` directory name since we seem to have rebranded our offering 6 times in the last 2 years (quoting Cody). This directory structure defends against the name churn.

codycushing:
@briangustafson as we agreed previously we should keep the service group name in the file since it is also present in the data source/resource name (I do much prefer the shorter names though).

@AlaaShaker @alexng-canuck We should evaluate moving the acceptance tests to a dedicated folder when we do the datasource/resource test merge.

@shalka data source and resource share code, which is likely to get moved to a code file of just the entity name--with that eventuality in mind, this name scheme optimizes for grouping related code.

Issue #322, opened 2017-10-20
Default Route Table/Default Security List objects are always created and there's no way to update them


lucassrg:
### Terraform Version

<details>
Terraform v0.10.6
</details>

### OCI Provider Version

<details>
terraform-provider-oci 2.0.2
</details>

### Description:
Running either the [vcn_full](https://github.com/oracle/terraform-provider-oci/blob/master/docs/examples/networking/vcn_full/vcn_full.tf) example or the [template file](https://gist.github.com/lucassrg/9b97fb224cb4882d7db6b04a5b048ea8) I created, the `Default Security List for <vcn>` or `Default Route Table for <vcn>` objects are *always* created and there's no way to remove them. You can just read these objects since they are available as [data sources](https://github.com/oracle/terraform-provider-oci/blob/7be1a4d5a8cc7895f742e3cfae5db3f07d081204/docs/datasources/core/virtual_networks.md).

Having these objects created by default seems to be a default behavior from OCI since these objects are also created via OCI Console. However, through the UI, there's a possibility to modify them, and I could not find a way to do it via Terraform, which lead us to have always 2 security lists created in case you need to specify any additional security rule not available in the default security list. The same applies for the Route table resource.

briangustafson:
The creation of those default resources happens at the service level when creating a VCN, and this is not something that can currently be disabled. You bring up a good point that there is no good way to modify these default resources from Terraform. This is something that we are tracking and will be looking into.

alexng-canuck:
The fix for this has been merged. Please see https://github.com/oracle/terraform-provider-oci/blob/master/docs/Managing%20Default%20Resources.md for details on how to configure default VCN resources in Terraform.

PR #374, opened 2017-11-29
Allow configuration of default resources under VCNs

alexng-canuck:
This addresses issue #322. Creating a VCN typically generates a
default DHCP option, route table, and security list resource. Allow
configuration of these default resources in the same way that they
are currently configured for non-default resources. Default
resources are identified via a new "default_id" field that marks
them as such.

To support this, we make following changes:
- Change Create, Get, Delete request handling for default resources
- Only allow compartment_id/vcn_id or default_id, but not both
- Add unit tests for default resources
- Add example TF file for configuring default VCN resources

One issue we may want to fix after this change:
- If we try to remove an updated default resource, nothing will happen to it and it retains its updated settings. The user may expect this to revert it to default settings. We can consider implementing this logic instead.

### DIFF
--- a/docs/examples/networking/vcn_default/vcn.tf
+++ b/docs/examples/networking/vcn_default/vcn.tf
@@ -1,0 +1,128 @@
+variable "tenancy_ocid" {}
+variable "user_ocid" {}
+variable "fingerprint" {}
+variable "private_key_path" {}
+variable "compartment_ocid" {}
+variable "region" {}
+
+provider "oci" {
+  tenancy_ocid = "${var.tenancy_ocid}"
+  user_ocid = "${var.user_ocid}"
+  fingerprint = "${var.fingerprint}"
+  private_key_path = "${var.private_key_path}"
+  region = "${var.region}"
+}
+
+resource "oci_core_virtual_network" "tf-vcn1" {
+  cidr_block = "10.0.0.0/16"
+  dns_label = "vcn1"
+  compartment_id = "${var.compartment_ocid}"
+  display_name = "tf-vcn1"
+}
+
+resource "oci_core_internet_gateway" "tf-ig1" {
+  compartment_id = "${var.compartment_ocid}"
+  display_name = "tf-ig1"
+  vcn_id = "${oci_core_virtual_network.tf-vcn1.id}"
+}
+
+resource "oci_core_default_route_table" "tf-default-route-table" {
+  manage_default_resource_id = "${oci_core_virtual_network.tf-vcn1.default_route_table_id}"
+  display_name = "tf-default-route-table"
+  route_rules {
+    cidr_block = "0.0.0.0/0"
+    network_entity_id = "${oci_core_internet_gateway.tf-ig1.id}"
+  }
+}
+
+resource "oci_core_route_table" "tf-route-table1" {
+  compartment_id = "${var.compartment_ocid}"
+  vcn_id = "${oci_core_virtual_network.tf-vcn1.id}"
+  display_name = "tf-route-table1"
+  route_rules {
+    cidr_block = "0.0.0.0/0"
+    network_entity_id = "${oci_core_internet_gateway.tf-ig1.id}"
+  }
+}
+
+resource "oci_core_default_dhcp_options" "tf-default-dhcp-options" {
+  manage_default_resource_id = "${oci_core_virtual_network.tf-vcn1.default_dhcp_options_id}"
+  display_name = "tf-default-dhcp-options"
+
+  // required
+  options {
+    type = "DomainNameServer"
+    server_type = "VcnLocalPlusInternet"
+  }
+
+  // optional
+  options {
+    type = "SearchDomain"
+    search_domain_names = [ "abc.com" ]
+  }
+}
+
+resource "oci_core_dhcp_options" "tf-dhcp-options1" {
+  compartment_id = "${var.compartment_ocid}"
+  vcn_id = "${oci_core_virtual_network.tf-vcn1.id}"
+  display_name = "tf-dhcp-options1"
+
+  // required
+  options {
+    type = "DomainNameServer"
+    server_type = "VcnLocalPlusInternet"
+  }
+
+  // optional
+  options {
+    type = "SearchDomain"
+    search_domain_names = [ "test123.com" ]
+  }
+}
+
+resource "oci_core_default_security_list" "tf-default-security-list" {
+  manage_default_resource_id = "${oci_core_virtual_network.tf-vcn1.default_security_list_id}"
+  display_name = "tf-default-security-list"
+
+  // allow outbound tcp traffic on all ports
+  egress_security_rules {
+    destination = "0.0.0.0/0"
+    protocol = "6"
+  }
+
+  // allow outbound udp traffic on a port range
+  egress_security_rules {
+    destination = "0.0.0.0/0"
+    protocol = "17" // udp
+    stateless = true
+
+    udp_options {
+      "min" = 319
+      "max" = 320
+    }
+  }
+
+  // allow inbound ssh traffic
+  ingress_security_rules {
+    protocol = "6" // tcp
+    source = "0.0.0.0/0"
+    stateless = false
+
+    tcp_options {
+      "min" = 22
+      "max" = 22
+    }
+  }
+
+  // allow inbound icmp traffic of a specific type
+  ingress_security_rules {
+    protocol  = 1
+    source    = "0.0.0.0/0"
+    stateless = true
+
+    icmp_options {
+      "type" = 3
+      "code" = 4
+    }
+  }
+}
--- a/docs/resources/core/dhcp_option.md
+++ b/docs/resources/core/dhcp_option.md
@@ -9,6 +9,8 @@ Provide a DHCP options resource.
 For more information, see
 [DNS in Your Virtual Cloud Network](https://docs.us-phoenix-1.oraclecloud.com/Content/Network/Concepts/dns.htm)
 
+For more information on configuring a VCN's default DHCP options, see [Managing Default VCN Resources](https://github.com/oracle/terraform-provider-oci/blob/master/docs/Managing%20Default%20Resources.md)
+
 ## Example Usage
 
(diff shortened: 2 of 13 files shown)


provider/core_route_table_resource.go
```diff
@@ -198,6 +222,15 @@ func (s *RouteTableResourceCrud) SetData() {
 }
 
 func (s *RouteTableResourceCrud) Delete() (e error) {
+	if s.D.Get("default_id") != "" {
+		// We can't actually delete a default resource.
+		// Instead, remove all the route rules first and mark it as deleted.
```
briangustafson: This is a good point. Should the fake delete operation reset the resource to its default state? I'll have to think this over a bit more. Is there any reason this is done for route rules but not for security lists and DHCP options?

alexng-canuck: For route rules, it's necessary to clear them to avoid dangling references to gateways. If we don't clear them, then "terraform destroy" will fake delete the route table and then try to delete the gateway. This would fail because the route rules are still referring to the gateway.

I'm leaning towards resetting the resource to its default state. For security list and DHCP options, it's a little more complicated because the default settings rely on values from the parent VCN resource.

One way we could handle this is to read and store the default state of the resource on Create, before we apply an updated settings. Then when we need to reset, just apply the stored state of the resource.

alexng-canuck:
Implemented logic for storing default config state at Create time and then restoring it upon Delete. Still missing tests for this logic, which would be a good idea to add.

alexng-canuck:
Latest commits address some feedback from @briangustafson 
- We rename the default_id attribute to reflect its role in modifying but not creating/deleting default resources
- Targeted deletion of a default resource now leaves the resource in an indeterminate state. Update the docs to reflect this behavior as well as how to generally configure default resources.

provider/core_dhcp_options_resource.go
```diff
@@ -139,6 +147,14 @@ func (s *DHCPOptionsResourceCrud) State() string {
 }
 
 func (s *DHCPOptionsResourceCrud) Create() (e error) {
+	// If we are creating a default resource, then don't have to
+	// actually create it. Just set the ID and update it.
+	if defaultId := s.D.Get("manage_default_resource_id").(string); defaultId != "" {
```
briangustafson: nit, but normally GetOk should be used to validate whether a property is set. The result should really be the same, but doesn't have to depend on the default value.

if vnicID, ok := s.D.GetOk("vnic_id"); ok {
		opts.VnicID = vnicID.(string)
	}

provider/core_dhcp_options_resource.go
```diff
@@ -154,6 +170,15 @@ func (s *DHCPOptionsResourceCrud) Get() (e error) {
 	res, e := s.Client.GetDHCPOptions(s.D.Id())
 	if e == nil {
 		s.Res = res
+
+		// If this is a default resource that we removed earlier, then
+		// we need to assume that the parent resource will remove it
+		// and notify terraform of it. Otherwise, terraform will
+		// see that the resource is still available and error out
+		if s.D.Get("manage_default_resource_id") != "" &&
```
briangustafson: What happens if I have a default resource in my config, do a targeted delete, and then apply again? Does that work correctly?

provider/core_security_list_resource_test.go
```diff
@@ -158,10 +242,19 @@ func (s *ResourceCoreSecurityListTestSuite) TestAccResourceCoreSecurityList_basi
...
+						display_name = "default-tf-security_list-updated"
+						egress_security_rules = []
+						ingress_security_rules = []
+					}
 				`,
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(s.ResourceName, "egress_security_rules.#", "0"),
 					resource.TestCheckResourceAttr(s.ResourceName, "ingress_security_rules.#", "0"),
+					resource.TestCheckResourceAttr(s.DefaultResourceName, "egress_security_rules.#", "0"),
+					resource.TestCheckResourceAttr(s.DefaultResourceName, "ingress_security_rules.#", "0"),
```
briangustafson: How about testing that a default resource can be applied, then removed from the config and applied, then added back and applied?

provider/core_dhcp_options_resource.go
alexng-canuck: I will address this. Should be easy to fix.

provider/core_dhcp_options_resource.go
alexng-canuck: Yes, when you target delete a default resource, it is removed from the Terraform state but still exists in OCI. If you then try to add it to the config again, this will result in a Terraform Create request that's converted to an Update API call by the provider.

provider/core_security_list_resource_test.go
alexng-canuck: This is a good idea for a test case. I will add this.

docs/examples/networking/vcn_default/vcn.tf
```diff
@@ -0,0 +1,128 @@
...
+  fingerprint = "${var.fingerprint}"
+  private_key_path = "${var.private_key_path}"
+  region = "${var.region}"
+}
+
+resource "oci_core_virtual_network" "vcn1" {
+  cidr_block = "10.0.0.0/16"
+  dns_label = "vcn1"
+  compartment_id = "${var.compartment_ocid}"
+  display_name = "vcn1"
```
parweza: nitpick: I like the naming convention of prepending the name with TF. It easily lets find the resources created from TF example

docs/resources/core/dhcp_option.md
```diff
@@ -59,8 +59,10 @@ resource "oci_core_dhcp_options" "dhcp-options2" {
 
 The following arguments are supported:
 
-* `compartment_id` - (Required) The OCID of the compartment.
-* `vcn_id` - (Required) The OCID of the VCN.
+* `compartment_id` - (Required) The OCID of the compartment. Do not specify with `manage_default_resource_id`.
```
parweza: isnt this optional then?

docs/resources/core/dhcp_option.md
```diff
@@ -59,8 +59,10 @@ resource "oci_core_dhcp_options" "dhcp-options2" {
...
+* `vcn_id` - (Required) The OCID of the VCN. Do not specify with `manage_default_resource_id`.
```
parweza: same as above

provider/core_dhcp_options_resource.go
```diff
@@ -139,6 +147,14 @@ func (s *DHCPOptionsResourceCrud) State() string {
 }
 
 func (s *DHCPOptionsResourceCrud) Create() (e error) {
+	// If we are creating a default resource, then don't have to
+	// actually create it. Just set the ID and update it.
+	if defaultId, ok := s.D.GetOk("manage_default_resource_id"); ok {
```
parweza: I dot see a validation that makes sure either (compartmentID, vcnId ) or defaultId is present.

provider/core_route_table_resource_test.go
```diff
@@ -14,13 +14,24 @@ import (
...
+	Provider            terraform.ResourceProvider
+	Providers           map[string]terraform.ResourceProvider
+	Config              string
+	ResourceName        string
+	DefaultResourceName string
 }
 
+var defaultRouteTable = `
+resource "oci_core_route_table" "default" {
+	manage_default_resource_id = "${oci_core_virtual_network.t.default_route_table_id}"
```
parweza: what is the expected behavior if user removes the manage default resource id after creating the resources

docs/resources/core/dhcp_option.md
alexng-canuck: You raise a good point. We need a way to say that you either must specify a compartment/vcn id or you must specify a default resource id. I will figure out a way to explain this.

provider/core_dhcp_options_resource.go
alexng-canuck: If both aren't present, the service API will fail with a message that the comparmentId/vcnId isn't valid.

That being said, I will add some validation logic here to catch this earlier.

provider/core_route_table_resource_test.go
alexng-canuck: This should fail with the same error as above.

alexng-canuck:
Discovered some edge cases that require hacks for making sure that the manage_default_resource_id and compartment_id/vcn_id don't conflict. Rather than chasing down this rabbit hole of unknown edge cases; I've decided to let Terraform do most of the schema validation for us before hand. 

In order to do this, I will create a new resource schema for default resources. Almost all of the logic remains the same; while having a new schema for default resources will let terraform validate the schema values for us properly. This simplifies our logic for handling default resource ID.

alexng-canuck:
Latest commits address edge cases and issues from @parweza feedback.

- Default resources now have their own resource schema but use the same underlying logic for CRUD operations. This avoids edge cases resulting from having two required parameter sets in compartment_id/vcn_id and manage_default_resource_id. This avoids the need for parameter validation logic as Terraform now does it for us via the default resource schema.

docs/resources/core/default_route_table.md
```diff
@@ -0,0 +1,42 @@
+# oci\_core\_default\_route\_table
```
briangustafson: Not blocking, but I like the idea of combining the default and regular docs for each resource better. That is, just add a line to each of the 3 resources about the default variant with pointer to the Managing%20Default%20Resources.md. Users looking to use the defaults will likely look at the regular resource first, before realizing that there's an entirely separate resource, and it cuts down on duplication.

Also, if we use separate docs then these need to be added to https://github.com/oracle/terraform-provider-oci/blob/master/docs/Table%20of%20Contents.md.

alexng-canuck: Valid feedback. I added the default resource docs for the sake of completeness, but I see that it's probably overkill and not easily discoverable. Will address this.

alexng-canuck:
Addressed documentation feedback from @briangustafson.

alexng-canuck:
Addressed feedback from @briangustafson, @rcohenma, and @codycushing  regarding how to handle removal of default resources.

PR #375, opened 2017-11-30
Add source port range to security list rules (rebased)

briangustafson:
The first commit in the PR is a port of #366 to the new directory structure. The second commit has some minor changes based on the comments from the PR.

Currently, TCP and UDP options can be specified with a single set of
max/min values, which correspond to the destination port range. This change
adds the ability to specify the source port range as well, issue #338.

Ideally, we should deprecate the existing max/min params and add a new
destination_port_range option next to the new source_port_range. Unfortunately,
a number of issue with nested schemas in Terraform are preventing support of both
of these params simultaneously. Instead, max/min params will remain, and source_port_range
will exist beside it.

This change also enables security lists with no rules.

Some updates are made to the docs, but the docs for the security list resource and data source
could really use a bigger update that is not included in this change.

### DIFF
--- a/provider/core_security_list_resource.go
+++ b/provider/core_security_list_resource.go
@@ -5,6 +5,7 @@ package provider
 import (
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/oracle/bmcs-go-sdk"
+
 	"github.com/oracle/terraform-provider-oci/crud"
 )
 
@@ -348,12 +349,13 @@ func buildPortRange(conf []interface{}) (portRange *baremetal.PortRange) {
 	if len(conf) > 0 && conf[0] != nil {
 		mapConf := conf[0].(map[string]interface{})
 
-		// The ok value will always be true, so we need to check against the default value instead.
 		max := mapConf["max"].(int)
 		min := mapConf["min"].(int)
 
 		// Max and Min default to 0, and that is not a valid port number, so we can assume that if
 		// the value is 0 then the user has not set the port number.
+		// Also, note that if either max or min is set, then the service will return an error if both are not
+		// set. However, we want to create the PortRange if either is set and let the service return the error.
 		if max != 0 || min != 0 {
 			portRange = &baremetal.PortRange{
 				Max: uint64(max),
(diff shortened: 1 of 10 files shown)


darkfibre:
I've tested this in our environment and it seems to be working as expected for both TCP and UDP source ports.

provider/core_security_list_resource.go
```diff
@@ -332,49 +322,94 @@ func (s *SecurityListResourceCrud) buildICMPOptions(conf map[string]interface{})
...
+
+	return
+}
+
+// Used to build rule lists for SetData.
+func buildConfRuleLists(res *baremetal.SecurityList) (confEgressRules, confIngressRules []map[string]interface{}) {
+	for _, egressRule := range res.EgressSecurityRules {
+		confEgressRule := map[string]interface{}{}
+		confEgressRule["destination"] = egressRule.Destination
+		confEgressRule = buildConfRule(
```
kohashim: The interface of the function `buildConfRule()` looks confusing because it takes a map (`confEgressRule` in this case) as input, then returns the map as it is, instead of a newly created map.  Probably `buildConfRule()` should create and return a new map as other `build*Options()` below.

provider/core_security_list_resource_test.go
```diff
@@ -35,16 +31,198 @@ func (s *ResourceCoreSecurityListTestSuite) SetupTest() {
 			cidr_block = "10.0.0.0/16"
 			compartment_id = "${var.compartment_id}"
 			display_name = "-tf-vcn"
+		}
+	    data "oci_core_security_lists" "t" {
```
kohashim: nit: indentation looks off.

provider/core_security_list_resource.go
```diff
@@ -332,49 +322,94 @@ func (s *SecurityListResourceCrud) buildICMPOptions(conf map[string]interface{})
...
-		confOpts := l[0].(map[string]interface{})
+	options := conf["tcp_options"].([]interface{})
+	if len(options) > 0 {
+		sourcePortRange, destinationPortRange := s.buildSourceAndDestinationPortRanges(options)
 		opts = &baremetal.TCPOptions{
-			baremetal.PortRange{
-				Max: uint64(confOpts["max"].(int)),
-				Min: uint64(confOpts["min"].(int)),
-			},
+			destinationPortRange,
```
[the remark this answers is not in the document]
briangustafson: sure, good suggestion

provider/core_security_list_resource.go
```diff
@@ -332,49 +322,94 @@ func (s *SecurityListResourceCrud) buildICMPOptions(conf map[string]interface{})
...
-		confOpts := l[0].(map[string]interface{})
+	options := conf["udp_options"].([]interface{})
+	if len(options) > 0 {
+		sourcePortRange, destinationPortRange := s.buildSourceAndDestinationPortRanges(options)
 		opts = &baremetal.UDPOptions{
-			baremetal.PortRange{
-				Max: uint64(confOpts["max"].(int)),
-				Min: uint64(confOpts["min"].(int)),
-			},
+			destinationPortRange,
```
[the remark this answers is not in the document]
briangustafson: done

provider/core_security_list_resource.go
briangustafson: done

provider/core_security_list_resource_test.go
briangustafson: I did a bunch of refactoring in this file in the latest revision in order to merge with the default resource changes... please take another look.

briangustafson:
Merged with default resource changes, now at #385.

PR #376, opened 2017-11-30
Add support for DBHomes

AlaaShaker:
Incorporates the generated support for the DBHomes resource.

The first commit is the actual change. The second commit contains the vendor folder changes which are also out for review (separate repo). Once I get the other changes published, I'll update this PR with another commit for the vendor folder _before_ pushing it.

IMPLEMENTS: https://jira.aka.lgl.grungy.us/browse/ORCH-438

### DIFF
--- a/docs/datasources/database/db_homes.md
+++ b/docs/datasources/database/db_homes.md
@@ -1,33 +1,38 @@
-# oci\_database\_db\_homes
+# oci\_database\_db_homes
 
-[DbHome Summary][40f7e502]
-
-  [40f7e502]: https://docs.us-phoenix-1.oraclecloud.com/api/#/en/database/20160918/DbHomeSummary/ "DbHomeSummary"
-
-Gets a list of database homes in the specified DB System and compartment.
+Gets a list of db_homes.
 
 ## Example Usage
 
 ```
-data "oci_database_db_homes" "t" {
-  compartment_id = "compartment_id"
-  db_system_id = "db_system_id"
-  limit = 1
-  page = "page"
+data "oci_database_db_homes" "testDBHomes" {
+	#Required
+	compartment_id = "${var.compartment_id}"
+	db_system_id = "${var.db_system_id}"
 }
+
 ```
 
-## Argument Reference
+## List Operation
+Gets a list of database homes in the specified DB System and compartment. A database home is a directory where Oracle database software is installed.
 
 The following arguments are supported:
 
-* `compartment_id` - (Required) The compartment OCID.
-* `db_system_id` - (Required) The OCID of the DB System.
-* `limit` - (Required) The maximum number of items to return.
-* `page` - (Optional) The pagination token to continue listing from.
+* `compartment_id` - (Required) The compartment [OCID](https://docs.us-phoenix-1.oraclecloud.com/Content/General/Concepts/identifiers.htm).
+* `db_system_id` - (Required) The [OCID](https://docs.us-phoenix-1.oraclecloud.com/Content/General/Concepts/identifiers.htm) of the DB System.
 
-## Attributes Reference
 
 The following attributes are exported:
 
-* `db_homes` - A list of database homes in the specified DB System and compartment.
+* `db_homes` - The list of db_homes.
+
+## DBHome Reference
+* `compartment_id` - The OCID of the compartment.
+* `db_system_id` - The OCID of the DB System.
+* `db_version` - The Oracle database version.
+* `display_name` - The user-provided name for the database home. It does not need to be unique.
+* `id` - The OCID of the database home.
+* `last_patch_history_entry_id` - The OCID of the last patch history. This is updated as soon as a patch operation is started.
+* `state` - The current state of the database home.
+* `time_created` - The date and time the database home was created.
+
--- a/docs/resources/database/db_home.md
+++ b/docs/resources/database/db_home.md
@@ -1,0 +1,65 @@
+# oci\_database\_db_home
+
+## Example Usage
+
+```
+resource "oci_database_db_home" "testDBHome" {
+	#Required
+	database {
+		#Required
+		admin_password = "${var.database_admin_password}"
+		db_name = "${var.database_db_name}"
+
+		#Optional
+		character_set = "${var.database_character_set}"
+		db_workload = "${var.database_db_workload}"
+		ncharacter_set = "${var.database_ncharacter_set}"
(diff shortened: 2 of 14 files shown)


docs/datasources/database/db_homes.md
```diff
@@ -1,33 +1,38 @@
-# oci\_database\_db\_homes
+# oci\_database\_db_homes
```
briangustafson: I assumed that we needed the slash, but it looks fine.

provider/database_db_homes_data_source.go
```diff
@@ -25,28 +26,20 @@ func DBHomesDatasource() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
-			"limit": {
```
briangustafson: Limit and page aren't very useful, but I don't think we can get rid of them. We should figure out a way to conditionally add these in the generator for existing resources and then I think we should mark them as deprecated.

provider/database_db_homes_resource.go
```diff
@@ -0,0 +1,243 @@
...
+			//Required
+			"database": {
+				Type:     schema.TypeList,
+				Required: true,
+				ForceNew: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						//Required
+						"admin_password": {
```
briangustafson: Sensitive: true,

provider/database_db_homes_resource.go
```diff
@@ -0,0 +1,243 @@
...
+		Update:   updateDBHome,
+		Delete:   deleteDBHome,
+		Schema: map[string]*schema.Schema{
+			//Required
+			"database": {
+				Type:     schema.TypeList,
+				Required: true,
+				ForceNew: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
```
briangustafson: This should probably just share the same schema as database_database_data_source.go (as should the database schema in database_db_system*.go). We should also be reading in service data about the database in SetData... and we might not want to use the full schema that includes purely computed fields until we do that.

provider/database_db_homes_resource.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func (s *DBHomeResourceCrud) Delete() (e error) {
+	dbHomeID := s.D.Get("id").(string)
+
+	e = s.Client.DeleteDBHome(dbHomeID, nil)
+	return
+}
+
+func (s *DBHomeResourceCrud) SetData() {
```
briangustafson: See https://github.com/oracle/terraform-provider-oci/blob/79fc22feeafb60a5951a8cc3c2c270615316e548/provider/core_instance_resource.go#L361 for an example of updating a nested resource. (It's the only example I'm aware of.) That also updates the nested resource.

provider/database_db_homes_resource_test.go
```diff
@@ -0,0 +1,216 @@
...
+		Steps: []resource.TestStep{
+			{
+				ImportState:       true,
+				ImportStateVerify: true,
+				Config:            s.Config,
+			},
+			{
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(s.ResourceName, "db_system_id", "newValue"),
```
briangustafson: Does this test work? Where is newValue set?

provider/database_db_homes_resource_test.go
```diff
@@ -0,0 +1,216 @@
+// Copyright (c) 2017, Oracle and/or its affiliates. All rights reserved.
```
briangustafson: I'm guessing that this takes a really long time to run. I'd be interested to discuss some improvements offline.

docs/resources/database/db_home.md
```diff
@@ -0,0 +1,65 @@
...
+* `db_system_id` - (Required) The OCID of the DB System.
+* `db_version` - (Required) A valid Oracle database version. To get a list of supported versions, use the [ListDbVersions](https://docs.us-phoenix-1.oraclecloud.com/api/#/en/database/20160918/DbVersion/ListDbVersions) operation.
+* `display_name` - (Optional) The user-provided name of the database home.
+
+
+## Update Operation
+Patches the specified dbHome.
+
+The following arguments support updates:
+* `db_version` - A valid Oracle database version. To get a list of supported versions, use the [ListDbVersions](https://docs.us-phoenix-1.oraclecloud.com/api/#/en/database/20160918/DbVersion/ListDbVersions) operation.
```
rcohenma: You need to remove this, as we won't support the update function. see the doc generator for how it should look

provider/database_db_homes_data_source.go
rcohenma: I would remove them because there is no valid use case for them

provider/database_db_homes_resource_test.go
rcohenma: newValue is meant to be replaced manually by an actual valid value

provider/database_db_homes_resource_test.go
```diff
@@ -0,0 +1,216 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(s.ResourceName, "display_name", "display_name"),
+					resource.TestCheckResourceAttrSet(s.ResourceName, "id"),
+				),
+			},
+		},
+	})
+}
+
+func (s ResourceDBHomeTestSuite) TestUpdateDBHomeDBVersion() {
```
rcohenma: need to remove all the update stuff. I don't think these tests can pass as written. Need to change all the "newValue" with actual valid values.

provider/database_db_homes_resource.go
rcohenma: change needs to be made in the generator

provider/database_db_homes_resource.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func (s *DBHomeResourceCrud) Get() (e error) {
+	dbHomeID := s.D.Get("id").(string)
+
+	s.Res, e = s.Client.GetDBHome(dbHomeID)
+	return
+}
+
+func (s *DBHomeResourceCrud) Update() (e error) {
```
rcohenma: remove

provider/database_db_homes_resource_test.go
rcohenma: should wait on test refactoring task of generator

PR #380, opened 2017-12-09
adding vcn id as part of import

parweza:
adding vcn id as part of import, fix for: https://github.com/oracle/terraform-provider-oci/issues/379

### DIFF
--- a/provider/core_internet_gateway_resource.go
+++ b/provider/core_internet_gateway_resource.go
@@ -171,6 +171,7 @@ func (s *InternetGatewayResourceCrud) SetData() {
 	s.D.Set("time_modified", s.Resource.ModifiedTime.String())
 	s.D.Set("state", s.Resource.State)
 	s.D.Set("time_created", s.Resource.TimeCreated.String())
+	s.D.Set("vcn_id", s.Resource.VcnID)
 }
 
 func (s *InternetGatewayResourceCrud) Delete() (e error) {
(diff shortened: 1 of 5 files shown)


provider/core_internet_gateway_resource.go
```diff
@@ -171,6 +171,7 @@ func (s *InternetGatewayResourceCrud) SetData() {
 	s.D.Set("time_modified", s.Resource.ModifiedTime.String())
 	s.D.Set("state", s.Resource.State)
 	s.D.Set("time_created", s.Resource.TimeCreated.String())
+	s.D.Set("vcn_id", s.Resource.VcnID)
```
briangustafson: Looks like this is missing in the data source as well.

provider/core_internet_gateway_resource_test.go
```diff
@@ -52,6 +52,7 @@ func (s *ResourceCoreInternetGatewayTestSuite) TestAccResourceCoreInternetGatewa
 					resource.TestCheckResourceAttrSet(s.ResourceName, "id"),
 					resource.TestCheckResourceAttrSet(s.ResourceName, "time_created"),
 					resource.TestCheckResourceAttrSet(s.ResourceName, "display_name"),
+					resource.TestCheckResourceAttrSet(s.ResourceName, "vcn_id"),
```
briangustafson: If you add this to the test but don't change provider/core_internet_gateway_resource.go, does this test still pass? I'm betting it does. (Either way you should add this - but generally I think we need to figure out a better testing strategy for import.)

provider/core_internet_gateway_resource.go
parweza: done

PR #388, opened 2017-12-13
Fix errors with terraform import because of missing vcn_id in tfstate files

rohanbhattacharjee:


vendor/github.com/oracle/bmcs-go-sdk/core_route_table.go
```diff
@@ -22,6 +22,7 @@ type RouteTable struct {
 	RouteRules    []RouteRule `json:"routeRules"`
 	State         string      `json:"lifecycleState"`
 	TimeCreated   Time        `json:"timeCreated"`
+	VcnID         string      `json:"vcnId"`
```
briangustafson: Should the resources for route table and internet gateway in the provider be updated to set the VCN ID as well?

rohanbhattacharjee: I know it's confusing, but the fix to the internet_gateway was started by @parweza - he has this PR (https://github.com/oracle/terraform-provider-oci/pull/380) that is waiting to be merged.

This PR has the fix for `route_table` and `dhcp_options` only.

RiaanRossouw:
Thanks for fixing this. I can confirm 2.0.6 it started working for me.  Prior to 2.0.6 I had to have an empty string ie vcn_id = "" for it to work.

PR #389, opened 2017-12-14
Fixes issue wherein policy was not destroyed and recreated when compartment is changed

rohanbhattacharjee:
…rtment was changed


briangustafson:
As discussed offline, ideally should be added to integration tests, but as long as it checks out manually I'm fine merging this first to get it out.

Issue #379, opened 2017-12-08
Difficulty importing network resources within VCN


darkfibre:
### Terraform Version

<details>
Terraform v0.11.1
</details>

### OCI Provider Version

<details>
2.0.4 - 2017-11-2
</details>

### Description:
I am having difficulties importing existing network resources that reside within a VCN, I'm not sure if this is an issue with the OCI provider or with Terraform itself. The import succeeds, but it looks like the import action does not associate the vcn_id with the imported resource so any subsequent terraform plan/apply wants to recreate everything.  Details / steps to reproduce:

#### Import the VCN-
<details>

```
# terraform import oci_core_virtual_network.test1 ocid1.vcn.oc1.iad.aaaaaaaazj46i3d75gkiirwsxb4wcwmnvfwhca55wqchxndkzl6j3wlwm4oa
oci_core_virtual_network.test1: Importing from ID "ocid1.vcn.oc1.iad.aaaaaaaazj46i3d75gkiirwsxb4wcwmnvfwhca55wqchxndkzl6j3wlwm4oa"...
oci_core_virtual_network.test1: Import complete!
  Imported oci_core_virtual_network (ID: ocid1.vcn.oc1.iad.aaaaaaaazj46i3d75gkiirwsxb4wcwmnvfwhca55wqchxndkzl6j3wlwm4oa)
oci_core_virtual_network.test1: Refreshing state... (ID: ocid1.vcn.oc1.iad.aaaaaaaazj46i3d75gkiirwsxb4wcwmnvfwhca55wqchxndkzl6j3wlwm4oa)

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```
</details>

#### Import the IG-
<details>

```
terraform import oci_core_internet_gateway.test1-IG ocid1.internetgateway.oc1.iad.aaaaaaaaypy65vi3pvdw7xrnhc5e3treq3bbrv3uemumyfrwc6pvv3m7pzua
oci_core_internet_gateway.test1-IG: Importing from ID "ocid1.internetgateway.oc1.iad.aaaaaaaaypy65vi3pvdw7xrnhc5e3treq3bbrv3uemumyfrwc6pvv3m7pzua"...
oci_core_internet_gateway.test1-IG: Import complete!
[...]

PR #393, opened 2017-12-15
Set "manage_default_resource_id" during import of default resources

alexng-canuck:
This fixes an issue where importing a default resource will leave
the manage_default_resource_id empty in the state file. Any change
in the TF config would force an unwanted delete/recreate.

Fix this by setting the manage_default_resource_id on import.

This should address issue with default resource imports described in #379

### DIFF
--- a/provider/core_dhcp_options_resource.go
+++ b/provider/core_dhcp_options_resource.go
@@ -12,7 +12,7 @@ import (
 func DefaultDHCPOptionsResource() *schema.Resource {
 	return &schema.Resource{
 		Importer: &schema.ResourceImporter{
-			State: schema.ImportStatePassthrough,
+			State: crud.ImportDefaultResource,
 		},
 		Timeouts: crud.DefaultTimeout,
 		Create:   createDHCPOptions,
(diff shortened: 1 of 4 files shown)


alexng-canuck:
@briangustafson , agree about having a test for importing default resources.

I tried looking for a way to pass the default resource ID into the ImportState tests. It's possible to implement an ImportStateIdFunc pointer that lets us read this value from the state. Unfortunately, this function was introduced recently and not in our terraform vendor code.

Let me know if you can think of another way to test imports.

provider/core_dhcp_options_resource.go
```diff
@@ -12,7 +12,7 @@ import (
 func DefaultDHCPOptionsResource() *schema.Resource {
 	return &schema.Resource{
 		Importer: &schema.ResourceImporter{
-			State: schema.ImportStatePassthrough,
+			State: crud.ImportDefaultResource,
```
rohanbhattacharjee: anyway to test these and assert that the "manage_default_resource_id" is set ?

In general, as Brian was saying the other day, our acceptance tests don't actually check for the import scenario passing. So, I am hoping you have manually verified that the state file is updated as you would expect it to.

PR #394, opened 2017-12-15
Forcing the terraform provider to use TLS 1.2 when communicating with the APIs

rohanbhattacharjee:


provider/provider.go
```diff
@@ -258,6 +258,10 @@ func ProviderConfig(d *schema.ResourceData) (clients interface{}, err error) {
 		)
 	}
 
+	clientOpts = append(clientOpts, baremetal.CustomTransport(
+		&http.Transport{TLSClientConfig: &tls.Config{MinVersion: tls.VersionTLS12, MaxVersion: tls.VersionTLS12}},
```
codycushing: Lets omit max version so as not to constrain to 1.2 for future tls versions:
>// MaxVersion contains the maximum SSL/TLS version that is acceptable.
// If zero, then the maximum version supported by this package is used,
// which is currently TLS 1.2.

rohanbhattacharjee:
@codycushing - updated the PR based on your comment

PR #397, opened 2017-12-19
Code for schema definitions made uniform

rohanbhattacharjee:
No change in functionality.

I just made use of the standard way we define schema for resources. Some of these files had a different code organization, which was causing trouble when parsing this code (for the code-gen comparison work).


provider/objectstorage_preauthrequest_resource.go
```diff
@@ -3,11 +3,11 @@
...
 import (
-	"github.com/hashicorp/terraform/helper/schema"
-	"github.com/oracle/bmcs-go-sdk"
-
 	"time"
 
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/oracle/bmcs-go-sdk"
 	"github.com/pkg/errors"
```
AlaaShaker: nit: Can you remove those empty lines between the github imports while you're at it :) I know it's not yours, but I saw them in a couple of places.

alexng-canuck:
Just a general comment that the parser shouldn't assume that nested schemas are always defined within their parent schema.

There are valid reasons to define a schema as a separate variable and to refer to the var inside of the parent schema (code reuse being one such reason). For an example, see "provider/core_security_list_resource.go", which defines a transport schema that is reused in multiple properties.

provider/objectstorage_preauthrequest_resource.go
rohanbhattacharjee: sure .. if i have to make another commit, I will take that ?

rohanbhattacharjee:
@alexng-canuck - if my parser has to work like a compiler or hunt out all possible ways of writing the same code, then it'll get really complicated, and take much longer time than what we'd like to spend on this task.

I spoke to @codycushing about this a few days back. He said, that according to him, this extracting out common schema definitions hasn't quite added much value. I agree - among some 90-odd resource/datasource files, 8 files have that different style of coding.

provider/objectstorage_preauthrequest_resource.go
codycushing: make fmt (goimports) will add space between standard go libs and 3rd party stuff, so we shouldn't fight it. It does fail at proper reformatting if there are odd new lines and custom import placement. If you want to see what it's supposed to look like according to goimports, collapse all empty lines and see what make fmt does

rohanbhattacharjee: I will keep an eye on this. Since Brian has approved this I'll go ahead and merge this.

Thanks guys.

codycushing:
Yeah, just because something can have the same schema does not mean it is the same schema. It's actually strange that a base schema worked at various points for identity, and the code to add un-shared properties defeated the purpose. The 1:1 approach is easier for generation as well. We've talked about a different base file approach to sharing schema between datasources and resources which me might employ with a larger refactor aided by generation in the future.

PR #398, opened 2017-12-22
Make examples testable

briangustafson:
Adds test cases TestExamplesPlan and TestExamplesApply, which will plan
and apply every example under /docs/examples. (There's an exception for
db_systems, which needs some work to make runnable using the stardard
set of variables.) The tests ensure that all examples are runnable using
the same set of variables.

Minor updates are made to the examples to get these tests to pass. There
are a lot more changes that should be made, but the first step is to get the 
tests to pass and then we can iterate.

In the future we might consider updating this to enable concurrent runs.


docs/examples/networking/route_table/route_table.tf
```diff
@@ -17,12 +13,25 @@ provider "oci" {
...
+  dns_label = "tfexamplevcn"
+}
+
+resource "oci_core_internet_gateway" "ExampleIG" {
+  compartment_id = "${var.compartment_ocid}"
+  display_name = "TFExampleIG"
+  vcn_id = "${oci_core_virtual_network.ExampleVCN.id}"
+}
+
+resource "oci_core_route_table" "ExampleRouteTable" {
```
alexng-canuck: Should we start making samples use the default route table and dhcp options? The VCN may introduce default DHCP option and security list settings that Terraform user is not aware of and isn't defined in their TF config. These defaults may also change in the future. Maybe we should make it a recommended practice to start defining the defaults so that nothing is left to chance.

briangustafson: I think that likely makes sense. Updating the examples and best practices will be the next step.

briangustafson:
@alexng-canuck - Yep, my original plan was to run each example as a TestStep, similar to our integration tests. Unfortunately, that framework does not seem to be well suited to running a real config spanning multiple files - if there's a good way to do it I didn't find it. The approach that's in this review does have the advantage of closely mirroring what an actual user would do.

provider/examples_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+const examples_test_state_file = "test_examples.tfstate"
+
+var examples_test_allowed_environment_variables = []string{
```
codycushing: these const and var names should be mixed caps instead of underscores

provider/examples_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"TF_VAR_user_ocid",
+	"TF_VAR_tenancy_ocid",
+	"TF_VAR_fingerprint",
+	"TF_VAR_private_key_path",
+	"TF_VAR_private_key_password",
+	"TF_VAR_region",
+	"TF_VAR_compartment_ocid",
+	"TF_VAR_AD",
+	"TF_VAR_ssh_public_key",
+	"TF_VAR_ssh_private_key",
```
codycushing: We can probably get rid of the confusing and only used in one example `TF_VAR_ssh_public_key` and `TF_VAR_ssh_private_key` environment variables using terraforms [file](https://www.terraform.io/docs/configuration/interpolation.html#file-path-) function

codycushing:
There is some inconsistent resource/variable renaming going on that would be good to resolve, example: `TerraformExampleUser`, `TFExampleVCN`, `ExampleVCN `, and `-tf-object`

provider/examples_test.go
briangustafson: Fixed

provider/examples_test.go
briangustafson: Probably, but let's do that in a later revision.

briangustafson:
@codycushing I pushed a couple commits to make the naming a bit more consistent. Generally, the naming in the examples is already inconsistent in a number of ways, and, while I don't want to this PR to make that significantly worse, it's also not meant to clean that up. Once we have this merged and can easily test the examples it will be much easier to do a larger refactor of the naming without fear of breaking things.

PR #402, opened 2018-01-02
Getting started guide

rohanbhattacharjee:


docs/Getting Started Tutorial.md
```diff
@@ -0,0 +1,267 @@
...
+
+Verify that your installation works as suggested in the linked page above.
+
+Run the following command -
+
+```bash
+terraform version
+```
+
+Ensure that the Terraform version is `v0.10.x` or higher.
```
briangustafson: Is this just so that we don't have to include the extra instructions for earlier versions?

docs/Getting Started Tutorial.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Install the OCI Terraform provider
```
briangustafson: A lot of this is duplicating info that's in the README. If this guide can offer a more streamlined, step by step process for this setup, or maybe something more specific whereas the general setup instructions are more general, then having instructions like this makes sense. As it is it doesn't seem to be different enough from the README instructions to warrant repeating.

docs/examples/getting_started/network.tf
```diff
@@ -0,0 +1,123 @@
...
+    protocol = "6"
+    source = "${var.VPC-CIDR}"
+  }]
+}
+
+# Creating a public subnet within our VCN.
+# Instances hosted in public-subnets can be accessed from the Internet.
+resource "oci_core_subnet" "MyPublicSubnetAD1" {
+  availability_domain = "${lookup(data.oci_identity_availability_domains.ADs.availability_domains[0], "name")}"
+  cidr_block = "10.0.1.0/24"
```
briangustafson: If we parameterize the VCN CIDR, then we should calculate the subnet CIDRs based on that using the cidrsubnet function - see https://www.terraform.io/docs/configuration/interpolation.html#cidrsubnet-iprange-newbits-netnum-

docs/examples/getting_started/network.tf
```diff
@@ -0,0 +1,123 @@
...
+    source = "0.0.0.0/0"
+  },
+    {
+      protocol = "6"
+      source = "${var.VPC-CIDR}"
+    }]
+}
+
+# The allow/deny rules in the security list determine if a subnet is public or private.
+# If we choose to run Oracle RDBMS in our private subnet, we might want to allow traffic to port 1521 etc.
```
briangustafson: This is rather specific for something that is not in this example.

AlaaShaker:
Do we need to run this be a technical writer?

PR #407, opened 2018-01-05
Enhancement requests...

nelsonse:
This update accomplishes several items:

1. More fully documents requirements for subnet security lists, AD specification format and exclusion of root compartment from use by this process.
2. Includes inline documentation in the configuration.tf file
3. Changes access method of RHEL ISO from OCI bucket/object method to URL. This allows any valid URL that contains the name of the ISO to be used in this process.
4. Changed the method of image selection for the ipxe instance to use the Operating System and OS Version vs. specifying the image name directly. Allows for selection of current images and ensures that deprecated images are not used.
5. Fixes issue when using on certain Linux distros where the fingerprint generation inserts an extraneous '(stdin)'
6. Removes the need for shar in generating the user-data script.


docs/solutions/rhel74_image/README.md
```diff
@@ -5,9 +5,10 @@ There are several prerequisites for using this process:
    RH Username and Password to allow you to temporarily subscribe the instance that is building the image 
    and get access to the various RH repos.
 2. The template expects pre-configured VCNs and Subnets.  
-3. You need to have a bucket in the OCI object store that contains the RHEL 7.4 iso.  The bucket must be in 
-   your tenancy, and must be accessible by the user executing the build.
+3. You need to provide a URL that points to the RHEL 7.4 ISO.  This URL must contain the name of the ISO, 
+   with an '.iso' extension.  An OCI PAR works well for this operation.
```
briangustafson: Might want to spell out pre-authenticated request and link to https://docs.us-phoenix-1.oraclecloud.com/Content/Object/Tasks/managingobjects.htm#par.

docs/solutions/rhel74_image/datasources.tf
```diff
@@ -35,20 +35,35 @@ data "oci_core_subnets" "subnet" {
 
 }
 
+# Gets the OCID of the image. This technique is for example purposes only. 
+# The results of oci_core_images may change over time for Oracle-provided images, 
+# so the only sure way to get the correct OCID is to supply it directly.
+
+data "oci_core_images" "image" {
```
briangustafson: Doesn't appear to be used anywhere. See https://github.com/oracle/terraform-provider-oci/pull/399/files, which switched to using the image OCID directly. That will be the recommended best practice going forward, so we should stick with that approach.

nelsonse: The problem with the image OCID is that it is specific to the region, so the user of the template must specify the correct OCID for each region.  This method (which is modified in the latest commit) allows the specification of the OS name and version, and selects the first one on the list.  This allows for the correct OCID to be specified regardless of the region.

The image datasource is used in the ipxe.tf to specify the image to be used for the ipxe instance.

New commit with both OS name and version forthcoming

docs/solutions/rhel74_image/README.md
nelsonse: Will do.  Will post when complete.

nelsonse: Completed.  Changes made.

docs/solutions/rhel74_image/datasources.tf
nelsonse: OS name and version in current commit.

docs/solutions/rhel74_image/variable.tf
```diff
@@ -22,25 +22,26 @@ variable "ipxe_instance" {
...
 }
 
-variable "region_all_zeros_ocid" {
+variable "ipxe_image_ocid" {
 	type = "map"
 	default = {
-		us-phoenix-1 = "ocid1.image.oc1.phx.aaaaaaaay27pdopotkapf2ahjlsn2wxndui5hn5w37hd2wss4ses4ol5xs6a"
-		us-ashburn-1 = "ocid1.image.oc1.iad.aaaaaaaaqftkoa5web2r7w4ls3wekgqmqy5f7untloetfiozyqbv2ql6qidq"
-		eu-frankfurt-1 = "ocid1.image.oc1.eu-frankfurt-1.aaaaaaaah4rggbyglst25peqd7vnyjzl6n5lwogiyllb6jaircakom46nswq"
+		us-phoenix-1 = "ocid1.image.oc1.phx.aaaaaaaaxklzl52nmabfp3466ilzfpo7lv737k44kih4jpo7nsmxjehwrdsq"
```
briangustafson: Looks like this is image "Oracle-Linux-7.4-2018.01.10-0" - would be useful to mention that in a comment.

docs/solutions/rhel74_image/variable.tf
```diff
@@ -22,25 +22,26 @@ variable "ipxe_instance" {
 	default = {
 		name = "ipxe-rhel74"
 		hostname = "ipxe-rhel74"
+		os = "Oracle Linux"
```
briangustafson: os and os-version aren't used, are they?

Issue #352, opened 2017-11-06
Results of oci_core_images will change over time for Oracle-provided images


briangustafson:
The data source "oci_core_images" is backed by the ListImages API, which returns both user-created and Oracle-provided images. However, the Oracle-provided images will change over time, with new images being added and old ones removed. That means that relying on this to get the OCID of an Oracle-provided image is likely to eventually break. However, hard-coding OCIDs is also problematic: the Oracle-provided images have different OCIDs for each region, which means that hard-coding an OCID will limit the config to a single region.

Currently, most of the examples in this provider filter images on display_name with the following note:

    # Gets the OCID of the image. This technique is for example purposes only. The results of oci_core_images may
    # change over time for Oracle-provided images, so the only sure way to get the correct OCID is to supply it directly.
    data "oci_core_images" "image-list" {
        compartment_id = "${var.compartment_ocid}"
        display_name = "Oracle-Linux-7.4-2017.10.25-0"
    }

This issue is tracking a better solution to this problem. (See related issues #342 and #345.)

Note that filters may also be used on oci_core_images, but that too will likely break with time for Oracle-provided images.

kohashim:
I think hard-coding OCIDs is a valid option here.  In fact, I believe most of users are hard-coding OCIDs for images for their production systems, rather than picking up a random image from ListImages API.  Thus the current examples with the data source "oci_core_images" show a technique which is not quite useful for actual use cases.

A map variable can be used for choosing a right OCID for each region.  The examples would need to be updated when a new region is added or the image needs to be updated for reasons such as a security update.

briangustafson:
Agree - using image OCIDs directly should be the standard approach for production use cases, and as you mention this could be selected from a map variable based on region. We'll update the docs and examples to reflect this.

briangustafson:
Most of the examples now reference the image OCID directly using a map variable - see for example https://github.com/oracle/terraform-provider-oci/blob/master/docs/examples/compute/private_ip/private_ip.tf#L66 .

parweza:
we are still using datasource in following:
docs/solutions/glusterfs/compute.tf
docs/solutions/omc_monitored_server/chef_stage_and_monitor_instance/main.tf
docs/solutions/omc_monitored_server/stage_and_monitor_instance/main.tf
docs/solutions/omc_monitored_server/stage_server/main.tf

rohanbhattacharjee:
[...]

PR #421, opened 2018-01-11
Use image OCID directly in all examples

briangustafson:
See issue #352. This switches the examples to the recommended method.
Some solutions still need to be updated.

### DIFF
--- a/docs/examples/networking/hybrid_dns/env-vars
+++ b/docs/examples/networking/hybrid_dns/env-vars
@@ -20,7 +20,7 @@ export TF_VAR_private_subnet_cidr=""
 
 ### Instance configuration
 export TF_VAR_InstanceShape=""
-export TF_VAR_InstanceImageDisplayName=""
+export TF_VAR_InstanceImageOCID=""
 
 ### Instance credentials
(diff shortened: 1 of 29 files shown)


docs/examples/networking/hybrid_dns/env-vars
```diff
@@ -20,7 +20,7 @@ export TF_VAR_private_subnet_cidr=""
 
 ### Instance configuration
 export TF_VAR_InstanceShape=""
-export TF_VAR_InstanceImageDisplayName=""
+export TF_VAR_InstanceImageOCID=""
```
alexng-canuck: What is the format if I were to provide this as an environment variable? Would this be a map?

alexng-canuck:
LGTM. Did have a question about how the image map format if specified as an env-var value?

docs/examples/networking/hybrid_dns/env-vars
briangustafson: Good question... it is a bit cumbersome:

  export TF_VAR_InstanceImageOCID='{us-phoenix-1="ocid1.image.oc1.phx.aaaaaaaalfnxla4wjqnkehn44xzu74orzdo5pxtsytjfyry72h4n5zrfmwoq"}'

PR #427, opened 2018-01-16
Add BACKLOG.md to track user-reported enhancement requests

alexng-canuck:
This will help us declutter the github issues list and allow it to
focus on bugs and blocking issues. This also gives us a way to
uniformly track and provide updates on these requested enhancements.

Enhancements coming from github issues list can be closed, once
they are tracked here.

### DIFF
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -1,0 +1,29 @@
+# Backlog
+This backlog is used to acknowledge and provide updates for community-requested enhancements from Github.
+
+These can be feature requests or improvements that address configurability, missing use cases, troubleshooting, and other known issues.
+
+## Planned
+
+Enhancement | Status | ETA
+--- | --- | ---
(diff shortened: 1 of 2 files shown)


codycushing:
This structure looks mostly fine, I dont think we should include an eta column. 

I believe this is something we should manage in a wiki page in this repo so that updates don't require the PR and commit process.

alexng-canuck:
@codycushing Fair point. Had this PR ready before realizing that we could just copy this over to the wiki.

anandhmahal:
@alexng-canuck - It would be nice if you can add a column for the GitHub issues ID. It would be easy to compare the IDs as opposed to the descriptions, when we cross check.

alexng-canuck:
Moving this backlog into the wiki instead: https://github.com/oracle/terraform-provider-oci/wiki/Feature-Request-and-Enhancements-Backlog

Issue #424, opened 2018-01-16
Renaming Object store bucket name in existing deployment is failing


sanketdjain:
### Terraform Version

Terraform v0.11.1

    provider.null v1.0.0
    provider.oci v2.0.5
    provider.template v1.0.0

### OCI Provider Version

    oci_v2.0.5 
    2018/01/05 09:35:28 [INFO] terraform-provider-oci 2.0.5

### Description:

1. Create a bucket in Terraform
2. Update the bucket name and apply again.

Terraform detects that the bucket name has changed, and it tries to **update** the resource. However, I don't think renaming buckets is supported in OCI and Terraform keeps trying again continuously. Eventually it fails with timeout.

I think OCI Plugin should detect it as a delete + create, if renaming is not supported, and the bucket is empty.

### Test Case:

```
// Backup buckets for CTRL
resource "oci_objectstorage_bucket" "bt_bkp_ctrl" {
  count = "${local.create_bucket ? var.ctrl_count : 0}"
  compartment_id = "${oci_identity_compartment.comp_backups.id}"
  name = "backup_${replace(element(oci_core_instance.serv_ctrl.*.hostname_label, count.index), "-", "")}"
  access_type = "ObjectRead"
  namespace = "${data.oci_objectstorage_namespace.nmsp.namespace}"
}
```

### Terraform Plan

Terraform will perform the following actions:

$ terraform plan
```
 <= data.oci_core_images.img
      id:             <computed>
      compartment_id: "ocid1.compartment.oc1..aaaaaaaanw5tvfobn7gujvbgkrcwlstuoczledvjcj6s3zchdp6burchb2fq"
      display_name:   "adwc_base_img"
      images.#:       <computed>

  ~ oci_objectstorage_bucket.bt_bkp_ctrl
      name:           "BACKUP_CTRL1" => "backup_ctrl1"

Plan: 0 to add, 1 to change, 0 to destroy.
```

kohashim:
Thank you for the report.  Confirmed the issue by reproducing.  A fix is now under review.

codycushing:
Fixed in https://github.com/oracle/terraform-provider-oci/releases/tag/2.0.7

PR #428, opened 2018-01-17
Recreate an object storage bucket when the name is changed

kohashim:
See the issue #424.  When the name of a bucket, the bucket and its containing objects are destroyed and created.  If the bucket contains objects managed outside terraform, the operation will fail due to 409 conflict error from the service API.

### DIFF
--- a/provider/objectstorage_bucket_resource_test.go
+++ b/provider/objectstorage_bucket_resource_test.go
@@ -37,6 +37,7 @@ func (s *ResourceObjectstorageBucketTestSuite) TestAccResourceObjectstorageBucke
 	resource.Test(s.T(), resource.TestCase{
 		Providers: s.Providers,
 		Steps: []resource.TestStep{
+			// test create
 			{
 				ImportState:       true,
 				ImportStateVerify: true,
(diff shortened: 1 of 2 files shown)


kohashim:
Thank you for the review.  I confirmed with the object storage service team (and with their code) that you cannot change the name of a bucket today.  The bucket name in UpdateBucketDetails object simply needs to match the one in the API REST URL.  Don't ask me why, that's just the way their code works. :)

provider/objectstorage_bucket_resource_test.go
```diff
@@ -56,6 +57,24 @@ func (s *ResourceObjectstorageBucketTestSuite) TestAccResourceObjectstorageBucke
 					resource.TestCheckResourceAttr(s.ResourceName, "metadata.content-type", "text/plain"),
 				),
 			},
+			// test update
+			{
+				Config: s.Config + tokenFn(`
+				resource "oci_objectstorage_bucket" "t" {
+					namespace = "${data.oci_objectstorage_namespace.t.namespace}"
+					compartment_id = "${var.compartment_id}"
+					name = "{{.token}}-changed"
```
kohashim: Does anyone know why "token" is used in the original test case (see the line 48 above)?  I kept using it here just in case, but I don't understand the intention why the bucket name needs to be dynamically generated for every run.

codycushing: It needs to be unique per test to prevent collisions with parallel tests or buckets that were not cleaned up from previous tests (if they had failed). It is `tokenize()` above that generates a new token and corresponding `tokenFn`. `tokenFn` just makes sure the same token is injected into each template string across multiple steps.

kohashim: Thanks Cody for the clarification!

PR #454, opened 2018-02-10
Attribute Reference

EscVector:
Switched Attribute Reference from virtual_networks to volume_attachments.


EscVector:
Should have been more specific in in comments... 
virtual_networks is incorrectly referenced in the volume_attachments documentation.  
Example should lead to  "data.oci_core_volume_attachments.cva.volume_attachments" not "data.oci_core_volume_attachments.cva.virtual_networks"

PR #480, opened 2018-03-15
Update instance examples to show interpolation syntax

alexng-canuck:
This shows usage of count variable and math operations to set-up
instances that can be attached to multiple block storage volumes.

### DIFF
--- a/docs/examples/compute/instance/block.tf
+++ b/docs/examples/compute/instance/block.tf
@@ -1,14 +1,16 @@
-resource "oci_core_volume" "TFBlock0" {
+resource "oci_core_volume" "TFBlock" {
+  count = "${var.NumInstances * var.NumVolumesPerInstance}"
   availability_domain = "${lookup(data.oci_identity_availability_domains.ADs.availability_domains[var.AD - 1],"name")}"
   compartment_id = "${var.compartment_ocid}"
-  display_name = "TFBlock0"
+  display_name = "TFBlock${count.index}"
   size_in_gbs = "${var.DBSize}"
 }
 
-resource "oci_core_volume_attachment" "TFBlock0Attach" {
+resource "oci_core_volume_attachment" "TFBlockAttach" {
+    count = "${var.NumInstances * var.NumVolumesPerInstance}"
     attachment_type = "iscsi"
     compartment_id = "${var.compartment_ocid}"
-    instance_id = "${oci_core_instance.TFInstance.id}"
-    volume_id = "${oci_core_volume.TFBlock0.id}"
+    instance_id = "${oci_core_instance.TFInstance.*.id[count.index / var.NumVolumesPerInstance]}"
+    volume_id = "${oci_core_volume.TFBlock.*.id[count.index]}"
 }
 
--- a/docs/examples/compute/instance/compute.tf
+++ b/docs/examples/compute/instance/compute.tf
@@ -1,7 +1,8 @@
 resource "oci_core_instance" "TFInstance" {
+  count = "${var.NumInstances}"
   availability_domain = "${lookup(data.oci_identity_availability_domains.ADs.availability_domains[var.AD - 1],"name")}"
   compartment_id = "${var.compartment_ocid}"
-  display_name = "TFInstance"
+  display_name = "TFInstance${count.index}"
   image = "${var.InstanceImageOCID[var.region]}"
   shape = "${var.InstanceShape}"
 
@@ -9,7 +10,7 @@ resource "oci_core_instance" "TFInstance" {
     subnet_id = "${oci_core_subnet.ExampleSubnet.id}"
     display_name = "primaryvnic"
     assign_public_ip = true
-    hostname_label = "tfexampleinstance"
+    hostname_label = "tfexampleinstance${count.index}"
   },
 
   metadata {
--- a/docs/examples/compute/instance/remote-exec.tf
+++ b/docs/examples/compute/instance/remote-exec.tf
@@ -1,19 +1,19 @@
 resource "null_resource" "remote-exec" {
-    depends_on = ["oci_core_instance.TFInstance","oci_core_volume_attachment.TFBlock0Attach"]
+    depends_on = ["oci_core_instance.TFInstance","oci_core_volume_attachment.TFBlockAttach"]
+    count = "${var.NumInstances * var.NumVolumesPerInstance}"
     provisioner "remote-exec" {
       connection {
         agent = false
         timeout = "30m"
-        host = "${data.oci_core_vnic.InstanceVnic.public_ip_address}"
+        host = "${oci_core_instance.TFInstance.*.public_ip[count.index % var.NumInstances]}"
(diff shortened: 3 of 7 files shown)


docs/examples/compute/instance/variables.tf
```diff
@@ -8,6 +8,14 @@ variable "compartment_ocid" {}
 variable "ssh_public_key" {}
 variable "ssh_private_key" {}
 
+variable "NumInstances" {
```
briangustafson: If I deploy with one set of values for these 2 variables and then change them, will it only create/destroy what is required, or will it destroy and recreate other resources?

alexng-canuck: Good question. The answer is it depends. If you change the NumInstances, then only the required resources will be created/destroyed.

If you modify the NumVolumesPerInstance, this could result in some volumes being reattached to other instances. I'm trying to think of a way to solve this. The interpolation syntax doesn't really have a way to track the previous mappings between volumes/instances.

docs/examples/compute/instance/compute.tf
```diff
@@ -1,15 +1,16 @@
 resource "oci_core_instance" "TFInstance" {
+  count = "${var.NumInstances}"
   availability_domain = "${lookup(data.oci_identity_availability_domains.ADs.availability_domains[var.AD - 1],"name")}"
   compartment_id = "${var.compartment_ocid}"
-  display_name = "TFInstance"
+  display_name = "TFInstance${count.index}"
```
AlaaShaker: Here, you're changing the main compute instance example from creating a single instance to n-instances. This is a general question: do we want a complex, mega-example that exercises a plethora of optional configurations (one compute example that does _everything_ you can do with an instance), or do we want targeted examples with overlap (single instance, n-instances, with metadata, with ipxeScript, etc.)? If we go with the latter, then you should leave this example as is & create a new one next to it.

I personally prefer multiple targeted examples with an overlap. They are scenario-based, easier to understand & target the different requirements & complexities the different users have. Plus, we already have a mega-example that exercises all the options (albeit, not runnable) in the docs.

alexng-canuck: I thought about this as well, and I agree that we should have multiple targeted examples with overlap (which is the point of this example). Really complex mega samples belong in the solutions folder.

The reason I decided to replace this particular example rather than creating a new one, is because the old example was pretty trivial and would've repeated 90% of the syntax of this example (i.e. this sample is a superset of what the old sample tried to demonstrate).

docs/examples/compute/instance/variables.tf
briangustafson: If you can solve it, great. It wouldn't surprise me if there isn't a good solution for this though, in which case maybe a comment would be good.

alexng-canuck: Yep. I will add a comment for this.

I couldn't find a good way to preserve the mappings with interpolation. Otherwise, you would need to manually duplicate the volumes and attachments to avoid using interpolation.

docs/examples/compute/instance/compute.tf
craigcarl-oracle: A create _n_ _x_ example is very much needed, this is a good place for it. I'd like to see most configurations use this model, even if _n_ = 1. The SQL Server Always On configuration uses this but it's a complicated configuration overall, that bit is hard to tease out.

docs/examples/compute/instance/block.tf
```diff
@@ -1,14 +1,16 @@
...
   compartment_id = "${var.compartment_ocid}"
-  display_name = "TFBlock0"
+  display_name = "TFBlock${count.index}"
   size_in_gbs = "${var.DBSize}"
 }
 
-resource "oci_core_volume_attachment" "TFBlock0Attach" {
+resource "oci_core_volume_attachment" "TFBlockAttach" {
+    count = "${var.NumInstances * var.NumVolumesPerInstance}"
     attachment_type = "iscsi"
```
craigcarl-oracle: If you change the attachment type to `paravirtualized` the volumes will be attached and usable when the customer logs in, it's a better experience.

docs/examples/compute/instance/remote-exec.tf
```diff
@@ -1,19 +1,19 @@
...
+        host = "${oci_core_instance.TFInstance.*.public_ip[count.index % var.NumInstances]}"
         user = "opc"
         private_key = "${var.ssh_private_key}"
     }
       inline = [
         "touch ~/IMadeAFile.Right.Here",
-        "sudo iscsiadm -m node -o new -T ${oci_core_volume_attachment.TFBlock0Attach.iqn} -p ${oci_core_volume_attachment.TFBlock0Attach.ipv4}:${oci_core_volume_attachment.TFBlock0Attach.port}",
-        "sudo iscsiadm -m node -o update -T ${oci_core_volume_attachment.TFBlock0Attach.iqn} -n node.startup -v automatic",
-        "echo sudo iscsiadm -m node -T ${oci_core_volume_attachment.TFBlock0Attach.iqn} -p ${oci_core_volume_attachment.TFBlock0Attach.ipv4}:${oci_core_volume_attachment.TFBlock0Attach.port} -l >> ~/.bashrc"
+        "sudo iscsiadm -m node -o new -T ${oci_core_volume_attachment.TFBlockAttach.*.iqn[count.index]} -p ${oci_core_volume_attachment.TFBlockAttach.*.ipv4[count.index]}:${oci_core_volume_attachment.TFBlockAttach.*.port[count.index]}",
```
craigcarl-oracle: All of this would go away with the change to `paravirtualized` volume attachments.

docs/examples/compute/instance/block.tf
alexng-canuck: The provider doesn't support paravirtualized volume attachments yet. Will add a note to add this once we have support for it.

alexng-canuck:
@craigcarl-oracle Could you give an example of what you mean by making the example 'modular'?

craigcarl-oracle:
A module is basically a function. They are stored outside of the configuration and are pulled in as required. They allow multiple configurations to reference the same resource in different ways. The SQL Server Always on configuration uses them - https://github.com/nelsonse/terraform-provider-oci/tree/master/docs/solutions/sql_server_always_on_availability_groups. The model is fairly new but it's the right way to do it. Eventually we will have a module for most of our resources here - https://registry.terraform.io/browse, customers can pull them from there programmatically with `terraform get`.

alexng-canuck:
@craigcarl-oracle Thanks for the explanation. I think we can convert the example to be modular at a later time (will file a github issue for this). The more pressing issue for now is that we have customers who are unaware of how 'count' can be used; and so getting this out will be tremendously helpful.

craigcarl-oracle:
@alexng-canuck Excellent plan, thanks!

docs/examples/compute/instance/variables.tf
alexng-canuck: Addressed in latest update.

PR #513, opened 2018-04-27
Fixed missing id field in example and added HCL syntax hightlight

anandhmahal:


rohanbhattacharjee:
As an internal dev .. can you possibly do this change in the bitbucket?

PR #514, opened 2018-04-27
Update env-vars

alistar79:
var miss match between env-vars and variables.tf

### DIFF
--- a/docs/examples/compute/instance/env-vars
+++ b/docs/examples/compute/instance/env-vars
@@ -11,7 +11,7 @@ export TF_VAR_region="<region in which to operate, example: us-ashburn-1, us-pho
 export TF_VAR_compartment_ocid="<compartment OCID>"
 
 ### Public/private keys used on the instance
-export TF_VAR_ssh_public_key_path=<path to public key>
-export TF_VAR_ssh_private_key_path=<path to private key>
+export TF_VAR_ssh_public_key=<path to public key>
+export TF_VAR_ssh_private_key=<path to private key>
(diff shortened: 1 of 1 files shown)


rohanbhattacharjee:
We have some inconsistencies in the examples in how certain variables are read and handled.
Like we found with the ssh key, some examples expect the content of the key, some others expect the key path is passed in.
compartment_id vs. compartment_ocid is another one that is inconsistent.

I have logged in internal bug to decide on what we prefer, and make all the examples consistent with it (instead of doing it one example at a time).

docs/examples/compute/instance/env-vars
```diff
@@ -11,7 +11,7 @@ export TF_VAR_region="<region in which to operate, example: us-ashburn-1, us-pho
 export TF_VAR_compartment_ocid="<compartment OCID>"
 
 ### Public/private keys used on the instance
-export TF_VAR_ssh_public_key_path=<path to public key>
-export TF_VAR_ssh_private_key_path=<path to private key>
+export TF_VAR_ssh_public_key=<path to public key>
```
rcohenma: "<path to public key>" should also be changed to "<public key>"
same for the private key

rcohenma: It seems we have a change in progress that will take care of this. It will be out soon.

briangustafson:
Thanks again for pointing out this issue. I see that this has since been addressed by a separate change at https://github.com/oracle/terraform-provider-oci/commit/c27090090dfab0a6e5d28eb8974b1809337d718f, so closing this PR.

PR #515, opened 2018-04-27
Update image uuid and the pdf referenced in variables.tf

alistar79:
outdated uuid for images


rohanbhattacharjee:
We just found out that Compute team now provides the OCIDs in a new page. It looks like the OCIDs have been updated (like you detected). In addition, the LHR image OCID also has a `-` in `london-1` ...

We are working to update all our examples en-mass to reflect this. This update would happen soon. Hence I would suggest we don't do this as a one-off change.

alistar79:
Cool thanks!

PR #516, opened 2018-04-27
Update env-vars

alistar79:
Needs the content of the pub and key for ssh not the filename of path as previous.  Worked this out from other examples

### DIFF
--- a/docs/examples/compute/instance/env-vars
+++ b/docs/examples/compute/instance/env-vars
@@ -11,7 +11,7 @@ export TF_VAR_region="<region in which to operate, example: us-ashburn-1, us-pho
 export TF_VAR_compartment_ocid="<compartment OCID>"
 
 ### Public/private keys used on the instance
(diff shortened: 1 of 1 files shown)


rohanbhattacharjee:
We have some inconsistencies in the examples in how certain variables are read and handled.
Like we found with the ssh key contents are expected by examples, sometimes the key path is expected.
`compartment_id` vs. `compartment_ocid` is another one that is inconsistent.

I have logged in internal bug to decide on what we prefer, and make all the examples consistent with it (instead of doing it one example at a time).

tsella:
I renamed the two variables as well, but changed the read to have **file()** instead of cat'ing into variable:

remote-exec.tf:        private_key = "${**file(var.ssh_private_key)**}"
compute.tf:    ssh_authorized_keys = "${**file(var.ssh_public_key)**}"

briangustafson:
Thanks again for pointing out this issue. I see that this has since been addressed by a separate change at c270900, so closing this PR.

PR #524, opened 2018-05-03
Hadoop Examples for Cloudera EDH

smithzc:
Included here are 4 template examples for Cloudera EDH on OCI.   This includes:

Sandbox - Single node running Cloudera Docker Container
Development - Small Beginner Cluster, no HA
Production Starter - Beginner Production Cluster with HA
N-Node - Scale with HA to any number of Workers

This also includes automation in post deployment leveraging bash and python to configure each cluster via the Cloudera Manager API.


EscVector:
Development Example:  Image in example doesn't appear accessible at least to my account.   I'm getting permission denied and have made sure I'm using Phoenix region.  Will it be published along with example?  Or what's an image to test with? 
 TF_VAR_image_ocid="ocid1.image.oc1.phx.aaaaaaaazs4djfpanpbh7zwqjqe4enku3juqo6dntkj66u2vjx4p2k2oj5cq"

Example:
* oci_core_instance.WorkerNode.0: Service error:NotAuthorizedOrNotFound. image   not found. http status code: 404
* oci_core_instance.WorkerNode[2]: 1 error(s) occurred:

smithzc:
Really I left that as an example - all of the fields in the env-vars need to be populated by the end user before deployment.   I owe this commit a readme.md to explain some of those bits. 

To answer your question though, the latest OEL 7.5 image for the targeted region should be used here: https://docs.us-phoenix-1.oraclecloud.com/images/image/1f7f90be-cb48-4a94-8940-9ddfe17fb5a9/

EscVector:
Great Stuff - Couple Notes for the README

- for dev, recommend just keeping admin/admin - saves a heck of a lot of time - everyone knows the default, or make the admin/admin replace an option.

- Line 70 remote_exec "/home/opc/.ssh/id_rsa" - would make a note to change that, or put as variable since it isn't picked up from validate and is I think the only place in the file that is referenced outside the versioned directory structure.

- Note that running demo as is requires 4T of Storage

- Requires sufficient resources limits to launch 5 images - I didn't have high enough limits to launch as as delivered, so changed to VM.Standard1.1, 1.4. and 1.8 shapes to test launch.  

Ultimately doing that may have resulted in limited resource because the deploy seemed to fail, but I've built 5 node clusters on 1.4 images as well as very underpowered VMs - Typical CDH BootCamp deploys 5 nodes, as example on moderately sized images.  

- Deploy Timing - Since deploy isn't InLine in Remote_Exec, the output screen returns before CM is ready to launch.  Would make a note on the timing and to not expect CM to be ready for a few minutes.  Seems simple, but could be confusing.  

Out of the gate the install looked like this with HDFS setup failing.
screen shot 2018-05-08 at 5 35 57 pm

screen shot 2018-05-08 at 5 52 40 pm

smithzc:
Thanks for the feedback!  Check your block storage config (if you're using Dev) - that's probably why the HDFS setup failed if it wasn't able to detect valid storage devices.   I have some script updates forthcoming specifically fixing Block provisioning logic - that has shown to be more problematic than the testing done with Bare Metal local storage.

EscVector:
Storage issues on boot - starting mid-April, we've had quite a few issues when spinning up images with subsequent remote-exec calls to spin up PeopleSoft. Previously, no issues and was really fast.  Thought it was storage at first, but seems the larger the image, the more latency just in the image start if starting from same image simultaneously, even when instance appears just fine and even after pam_nologin.  Figured it an issue with under the hood resource contention when pulling from the same image. The odd fix was to clone the image and launch instances from two separate, but identical images.  Storage was good, everything good, no code change except to start from two different images and that fixed a world of pain.

smithzc:
Moving this to a different location on oci-provider.

PR #534, opened 2018-05-23
Updates with new images and other fixes...

nelsonse:
What is fixed/changed in this release:

- Fixed issue with OCI-CLI and pip that was preventing the process from working.  The original 
  process attempted to install OCI-CLI as a global resource in the global python repo...this broke
  after some changes in pip.  So now we install pip locally to each user.
- Changed our blank image OCID to use a new global value.  This makes the process *more* portable across
  regions (but not completely...see next bullet).
- Updated the ipxe server image list to include the LON region.  LON is now fully supported on this process.
- *** IMPORTANT CHANGE *** Updated process to use Instance Principal authorization method for OCI-CLI.  This
  removes the requirement to upload the user private key into the ipxe server.  *HOWEVER*, it does REQUIRE 
  the creation of a Dynamic Group for the Compartment this is being executed in.  This is included in the
  prerequisites section below.  The new process WILL NOT WORK without an Instance Principal (also known
  as a Dynamic Group).


goltermann:
@nelsonse ack on the OCID changes.  Small nit, you should use "LHR", not "LON" since that's the region key for uk-london-1 https://docs.us-phoenix-1.oraclecloud.com/Content/General/Concepts/regions.htm

nelsonse:
Changed LON to LHR

craigcarl-oracle:
@nelsonse Please resolve the conflicts so we can merge this. Why is an update to `docs/solutions/sql_server_always_on_availability_groups/README.md` included in this pull?

PR #566, opened 2018-08-04
 Fix for multi DBS datasources output.

nirmalpathak:
This fix allows output to print all the IP Addresses using datasource if multiple DBS are created.


alexng-canuck:
@nirmalpathak , wondering if you had any response to the comments left for this PR?

nirmalpathak:
My changes works if I have modular code, e.g
```
module "dbs1" {
 source           = "../module/db"
}

module "dbs2" {
 source           = "../module/db"
}
```

It doesn't work if I have DB count in DBS resource, e.g.

```
resource "oci_database_db_system" "dbs" {
 count               = "2"
..
}
```

I want to printout something like this, when I create multiple DBS.

```
DBSHostnames = [
    dbs1,
    dbs2
]
DBSPrivateIPs = [
    172.16.0.2
]
```

PR #596, opened 2018-09-17
Case sensitive check for DBaaS lifecycle status

DimeCaifan:
line 37, the value for default should be "Available". as it appears that the field is case sensitive.

### DIFF
--- a/docs/examples/database/adw/variables.tf
+++ b/docs/examples/database/adw/variables.tf
@@ -34,5 +34,5 @@ variable "autonomous_data_warehouse_backup_display_name" {
 }
 
 variable "autonomous_data_warehouse_backup_state" {
-  default = "AVAILABLE"
+  default = "Available"
 }


alexng-canuck:
@ravitandonrt, @bowfeng: Can you review this PR and provide a justification of whether we should accept it?

ravitandonrt:
@alexng-canuck I ran tests locally and also followed up with the customer (who was internal).

The issue had occurred because the database went `Unavailable` and hence did not return when `AVAILABLE` was passed. However, once it had recovered, `Available` worked due to coincidence.

Lifecycle states in OCI are supposed to be all capital letters. Service can choose to accept both case sensitive or insensitive query parts, and in this case databases APIs seems to accept both.

For anyone reading this, this is an optional field for data sources, if you specify only databases that match that state will be returned, if you omit this field, all databases (ADWs) will be returned matching other filters specified.

The backups are a slight different story, only after ADW has been configured for backups, will that resource get created.

kilokahn:
@DimeCaifan I am going to close this PR as we don't really want to tie ourself to the service behavior as noted by @ravitandonrt

PR #627, opened 2018-10-12
Add version to access ADWC resource

loiclefevre:
Version set to 3.4. Without it, ADWC resources were not recognized.

### DIFF
--- a/docs/examples/database/adw/provider.tf
+++ b/docs/examples/database/adw/provider.tf
@@ -4,4 +4,5 @@ provider "oci" {
   fingerprint      = "${var.fingerprint}"
   private_key_path = "${var.private_key_path}"
   region           = "${var.region}"
+  version          = "3.4"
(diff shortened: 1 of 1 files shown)


ravitandonrt:
@loiclefevre - Thanks for taking the time to share your feedback.

The ADW resource was added in version 2.2.4 of OCI Terraform provider - https://github.com/terraform-providers/terraform-provider-oci/blob/master/CHANGELOG.md#224---2018-09-11

We bumped our version to 3.x once we moved into Hashicorp registry. This PR ties this example to just version 3.4 of terraform provider for OCI.

While this change may make sense locally for someone trying to be on latest provider version we released yesterday, I do not think we can accept this PR without more evidence - that is in what version was it not recognized?

Please file an issue with logs on the version information where it may not have worked in a 3.x release and we can fix the example appropriately.

PR #630, opened 2018-10-19
Fix website for the dns record

lulichn:
Fixed website for the oci_dns_records resource.

https://www.terraform.io/docs/providers/oci/r/dns_record.html


afedorch:
Thanks for the fix

PR #638, opened 2018-11-02
Fixing broken links

mbfrahry:
Hey all. Found a couple broken links that this PR fixes


ravitandonrt:
Thanks @mbfrahry for reporting these. I had recently updated our doc generation to handle all URL replacements but it didn't capture incomplete URLs, I will see if I can add a better rule to capture these as well.

Apart from ```make website-test``` is there any other tool you use to find these issues, any other target in master website repo that we can run to ensure everything is good as part of release process?

mbfrahry:
`make website-test` or `make website` is your best bet. It'll tell you exactly what links are broken

PR #698, opened 2019-01-25
Update tags.tf

rramge:
The cost tracking boolean flag has been missing from the example.


docs/examples/identity/tags.tf
```diff
@@ -12,6 +12,8 @@ resource "oci_identity_tag" "tag1" {
   description      = "tf example tag"
   name             = "tf-example-tag"
   tag_namespace_id = "${oci_identity_tag_namespace.tag-namespace1.id}"
+  is_cost_tracking = false          // the value "true" is only permitted if the associated tag namespace is part of the root compartment
```
afedorch: Thanks for the PR!
Could we remove redundant empty line and move `is_cost_tracking` attribute to the optional block?

PR #703, opened 2019-01-30
Update Oracle Copyright notices for 2019

ravitandonrt:
Updating the Copyright notices for all source files, scripts and examples to include initial released year 2017 through 2019.

Note: Some files only contain newline at the end to make all of them consistent.


ravitandonrt:
Thanks for pointing out @alexng-canuck, updated those scripts and GNUMakefile as well, and also changed copyright location after shebang.

PR #731, opened 2019-03-13
website: fix myriad broken links to https://docs.cloud.oracle.com/iaas/...etc

nfagerlund:
These are breaking the website build. If there's not a release imminent, it'll need cherry-picking to stable-website too.

### DIFF
--- a/website/docs/d/ons_notification_topic.html.markdown
+++ b/website/docs/d/ons_notification_topic.html.markdown
@@ -25,7 +25,7 @@ data "oci_ons_notification_topic" "test_notification_topic" {
 
 The following arguments are supported:
 
-* `topic_id` - (Required) The [OCID](/iaas/Content/General/Concepts/identifiers.htm) of the topic to retrieve. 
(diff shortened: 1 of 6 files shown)


kilokahn:
I have also cherry-picked this change to `stable-website`

nfagerlund:
@kilokahn Thanks so much! :D 

I'm looking through the Makefile in https://github.com/hashicorp/terraform-website right now, and I _think_ something like this would meet your needs. Assuming that you've arranged things so there's:

- A checkout of this repo at `/tmp/terraform-provider-oci`
- A checkout of the website repo at `/tmp/terraform-website`

...then, like...

```
$ cd /tmp/terraform-website
$ PROVIDER_PATH="/tmp/terraform-provider-oci" PROVIDER_NAME="oci" make website-provider-test
```

I haven't actually tried this, because I'm about chin-deep in eight other PRs much like this one right now. 😅 But hopefully that's useful for getting you started.

PR #801, opened 2019-06-25
Release v3.31.0 2

dshelbyo:
Added
- Support for moving `email sender` resource between compartments
- Support for moving NAT Gateway resource across Compartments

### Fixed
- Fix for `defined_tags` property deletion bug

## Notes
- This release upgrades the Terraform plugin SDK to v0.12.3-0.20190619193004-2ab2796c932c, which fixes 
how null/empty values are stored in state during import and fixes unnecessary diffs caused by omission of
Optional/Computed fields.


dshelbyo:
checks failed.

PR #816, opened 2019-07-05
Block storage limits were increased from 16TB to 32TB

RobertClarke:
Signed-off-by: Robert Clarke <robert.clarke@oracle.com>

Limits in OCI were increased last year: https://blogs.oracle.com/cloud-infrastructure/oracle-shatters-cloud-storage-limits-with-the-best-performance

### DIFF
--- a/vendor/github.com/oracle/oci-go-sdk/core/instance_configuration_instance_source_via_image_details.go
+++ b/vendor/github.com/oracle/oci-go-sdk/core/instance_configuration_instance_source_via_image_details.go
@@ -20,7 +20,7 @@ import (
 // InstanceConfigurationInstanceSourceViaImageDetails The representation of InstanceConfigurationInstanceSourceViaImageDetails
 type InstanceConfigurationInstanceSourceViaImageDetails struct {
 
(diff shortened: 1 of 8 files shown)


alexng-canuck:
@parrneet One PR you should be aware of. We may need to refresh our generated documentation to cover this.

kilokahn:
@RobertClarke The API docs at https://docs.cloud.oracle.com/iaas/api/#/en/iaas/20160918/datatypes/InstanceSourceViaImageDetails have also not been updated for the changed maximum limits. We will reach out to the product team to get it updated upstream, and the Terraform docs will get updated as a downstream effect. I am closing this PR since the docs will get auto-generated, and we don't want to have inconsistency in the docs across the different surfaces (SDK, CLI, Terraform, and Console).

kilokahn:
It is also unclear if the boot volume sizes have actually increased as well. From the blog content, it looks like it is pointing to the block volume sizes (and not necessarily boot volume sizes). Anyway, once the upstream documentation is updated, it should flow through anyway.

PR #822, opened 2019-07-16
Use correct value for can_use_auth_tokens

hatchan:
The value for can_use_api_keys could override the value for can_use_auth_tokens
and this could cause issues when planning a change.

Signed-off-by: Benno van den Berg <benno.van.den.berg@oracle.com>


hatchan:
Btw I was not able to run the test locally, but this is the same configuration that didn't work for me.

kilokahn:
```
--- PASS: TestResourceIdentityUserCapabilitiesManagementTestSuite (5.23s)
    --- PASS: TestResourceIdentityUserCapabilitiesManagementTestSuite/TestAccResourceIdentityUserCapabilitiesManagement_basic (5.23s)
PASS
ok      github.com/terraform-providers/terraform-provider-oci/oci       5.277s
```
I was able to verify that the test fails locally without the fix, and passes after the fix.

PR #917, opened 2019-11-04
Add subcategories to docs for Terraform Registry

pwelch:
We will soon be displaying documentation for this provider both on
    [terraform.io](https://www.terraform.io/docs/providers/index.html)
    and on [the Terraform Registry](https://registry.terraform.io/providers).

    Documentation displayed on the Terraform Registry will not use the ERB
    layout containing the existing navigation, and will instead build the
    navigation dynamically based on the directory and YAML frontmatter of a file.
    For providers that group similar resources and data sources by service or
    use case (subcategories), we'll need to add this information to the
    Markdown source file.

    This PR modifies Resource and Data Source documentation source files by
    adding a subcategory to the metadata if those files were grouped previously.

    For more information about how documentation is rendered on the
    Terraform Registry, please see this reference:
    [Terraform Registry - Provider Documentation](https://www.terraform.io/docs/registry/providers/docs.html).

### DIFF
--- a/scripts/docscheck.sh
+++ b/scripts/docscheck.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+docs=$(ls website/docs/**/*.markdown)
+error=false
+
+for doc in $docs; do
+  dirname=$(dirname "$doc")
+  category=$(basename "$dirname")
+
+
+  case "$category" in
+    "guides")
+      # Guides have no requirements
+      continue
+      ;;
+
+    "d")
+      # Data sources have no requirements
+      continue
+      ;;
+
+		"r")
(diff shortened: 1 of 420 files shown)


scripts/docscheck.sh
```diff
@@ -0,0 +1,44 @@
...
+
+
+  case "$category" in
+    "guides")
+      # Guides have no requirements
+      continue
+      ;;
+
+    "d")
+      # Data sources have no requirements
```
jotruon: @pwelch datasources do require the subcategory

pwelch: @jotruon Updated the script

PR #982, opened 2020-02-21
apigw docs improvement

abtris:


alexng-canuck:
Hello @abtris ,

Thanks for the pull request and feedback. As our documentation is generated from a single source for disparate tool sets, it will be easier to maintain these changes if we addressed this feedback in our internal documentation and updated the documentation ourselves.

PR #1033, opened 2020-04-17
Migrate to standalone plugin SDK Go module

### LINKED ISSUE #1001: Migrate to standalone Terraform plugin SDK
This provider depends on the `github.com/hashicorp/terraform` Go module, which has been superseded by a standalone Terraform Plugin SDK Go Module:

https://www.terraform.io/docs/extend/plugin-sdk.html

We recommend that the OCI provider migrate to the new plugin SDK in order to take advantage of new features and benefit from bug fixes.

## Requirements for migration

The above documentation describes how the provider can be migrated to use the new module automatically using the [`tf-sdk-migrator`](https://github.com/hashicorp/tf-sdk-migrator) tool. 

This provider meets all prerequisites for migration except for the following:

### `VersionString()`

`terraform.VersionString()` has now been deprecated. This identifier is used in:
    - `terraform-provider-oci/oci/provider.go:961:128`
    - `terraform-provider-oci/oci/provider_test.go:416:128`

The OCI provider includes `terraform.VersionString()` in the user agent for client requests. It also includes `terraformCLIVersion`, which is set to `ociProvider.TerraformVersion`. Prior to migrating to the standalone SDK, these values may differ, as they represent the version of Terraform that the provider is compiled against (`terraform.VersionString()`), and the version of the Terraform CLI that is being used at runtime (`ociProvider.TerraformVersion`) .

After migrating to the standalone SDK, `terraform.VersionString()` is no longer available, because the provider no longer imports Terraform Core. I recommend that the user agent include instead [`meta.SDKVersionString()`](https://github.com/hashicorp/terraform-plugin-sdk/blob/master/meta/meta.go#L31), the version of the plugin SDK that the provider is compiled against.

### Deprecated `terraform` packages

The following packages have been deprecated and are not available in the new SDK:
 - `github.com/hashicorp/terraform/backend/local`
 - `github.com/hashicorp/terraform/command`

These packages are used in the resource discovery/export functionality introduced in https://github.com/terraform-providers/terraform-provider-oci/pull/913, specifically in `export_compartment.go`. My understanding is that they are used for operations including manipulation of the Terraform state file, and running Terraform CLI commands.

#### Replacing deprecated packages

This extremely valuable feature is unusual for a Terraform provider, and I don't know of any others that instrument Terraform CLI in this way. The Terraform team is currently designing the next major versions of the plugin SDK, and we are actively looking into ways to support resource discovery and config generation for a wide range of providers.

At the moment, however, the plugin SDK does not have helper functions for running Terraform CLI commands. I believe it is possible to replace usage of the deprecated packages in `export_compartment.go` without the use of any other dependencies. For an example of how this might be done, please see the following library currently in use by the SDK: https://github.com/hashicorp/terraform-plugin-test/blob/master/terraform.go#L112
This library is not suitable for use directly by providers - the code is provided as an example only.

---

Please comment on this issue or reach out to the Terraform team if you have any questions about migrating to the plugin SDK. Once the above prerequisites are met, the migration should require only running `tf-sdk-migrator migrate`, and have no impact on end users.

kmoe:
Fixes #1001

This PR migrates this provider to the standalone Terraform Plugin SDK, v1.15.0.

It also makes the following changes, prerequisites for migration:

#### Replace deprecated function calls

The `github.com/hashicorp/terraform#VersionString()` function is not available in the standalone plugin SDK. This function returned a string representing the version of the Terraform Core codebase which the provider had been compiled against. It has been replaced with `github.com/hashicorp/terraform-plugin-sdk/meta#SDKVersionString()`, the equivalent function for the new plugin SDK.

#### Replace use of deprecated packages

The packages `github.comhashicorp/terraform/backend/local` and `github.comhashicorp/terraform/command` are deprecated for Terraform providers and are not available in the standalone plugin SDK module, `github.com/hashicorp/terraform-plugin-sdk`.

This PR removes these imports and reimplements the functionality using the `github.com/hashicorp/terraform-exec` module, an existing dependency of the standalone SDK. Currently, an error will be returned if this command cannot be found in the system PATH, but if required, the module can install and verify Terraform automatically, by changing `export_compartment.go:358` to the following:
```
tfPath, err := tfinstall.Find(tfinstall.LookPath(), tfinstall.LatestVersion(installDir, true))
```
where `installDir` is the directory where the binary should be downloaded.

**N.B. The export compartment feature has not been retested after this change.**


oci/provider.go
```diff
@@ -500,7 +501,7 @@ func buildConfigureClientFn(configProvider oci_common.ConfigurationProvider, htt
 		terraformCLIVersion = ociProvider.TerraformVersion
 	}
 	userAgentProviderName := getEnvSettingWithDefault(userAgentProviderNameEnv, defaultUserAgentProviderName)
-	userAgent := fmt.Sprintf(userAgentFormatter, oci_common.Version(), runtime.Version(), runtime.GOOS, runtime.GOARCH, terraform.VersionString(), terraformCLIVersion, userAgentProviderName, Version)
+	userAgent := fmt.Sprintf(userAgentFormatter, oci_common.Version(), runtime.Version(), runtime.GOOS, runtime.GOARCH, sdkMeta.SDKVersionString(), terraformCLIVersion, userAgentProviderName, Version)
```
paultyng, line 504: It may be easier to just drop the SDK version since its implied by the provider version.

parrneet:
Closing this PR, We will integrate the changes .

PR #1036, opened 2020-04-20
DNS Import Support

harveylowndes:
Adds support for DNS resource imports and updates the web markdown accordingly. Imports look like:
`terraform import 'module.my_module.oci_dns_record.test_record' 'zones/<zone>/records/<domain>/rtypes/<rtype>/recordHashes/<record hash>'`

optionally you can attach `compartmentId/<compartment id>` at the end. 

This pattern is open to discussion, I just followed what I have seen elsewhere for DNS in the API.

Signed-off-by: Harvey Lowndes <harvey.lowndes@oracle.com>

### DIFF
--- a/oci/dns_record_resource.go
+++ b/oci/dns_record_resource.go
@@ -12,6 +12,8 @@ package oci
 import (
 	"context"
 	"fmt"
+	"net/url"
+	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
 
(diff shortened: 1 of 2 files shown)


jotruon:
Hi @harveylowndes,
I do some testing and this change cause a breaking change. 
The new composite ID will break the old ID.

I need to investigate more to support import for DNS record.

PR #1101, opened 2020-06-26
Corrected docs for object storage to mark content field as optional

srishtipmishra:
Corrected docs for object storage to mark content field as optional


srishtipmishra:
Abandoning this PR.

Issue #1191, opened 2020-10-05
oci_objectstorage_object Error: Service error:ContentLengthRequired. The Content-Length header is required. http status code: 411. Opc request id: iad-1:bQGujHV5


tutorialbyexample:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version and Provider Version
terraform -v
Terraform v0.12.19
+ provider.oci v3.59.0

### Affected Resource(s)

```
resource "oci_objectstorage_object" "objectstorage_object" {
    for_each = var.oci_objectstorage_object
    #Required
    bucket                  = var.object_bucket_name
    content                 = each.value.content
    namespace               = var.object_namespace
    object                  = each.value.object
    
    #Optional
    cache_control           = each.value.cache_control
    content_disposition     = each.value.content_disposition
    content_encoding        = each.value.content_encoding
    content_language        = each.value.content_language
    content_type            = each.value.content_type
    metadata                = each.value.metadata
}
oci_objectstorage_object = {
      "iadt-object" = { bucket_name = "iadt-bucket",
                            content = "", 
                            object = "iadt-object", 
                            cache_control = "", 
                            content_disposition = "", 
                            content_encoding = "", 
                            content_language = "en-US", 
                            content_type = "application/json", 
                            metadata = {}}
 }

```

### Terraform Configuration Files

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. 
# Please remove any sensitive information from configuration files before sharing them. 
```

### Expected Behavior
Working in us-phoenix-1 region with same code and same variable value but not working in IAD.

### Actual Behavior
```
Error: Service error:ContentLengthRequired. The Content-Length header is required. http status code: 411. Opc request id: iad-1:WsnJbwfNJUgtTwH-_gWhn5q_zVqVCJs6XgLuOXscJN6tKb3iz-sCpYBTQiQZjLWl

  on .terraform/modules/objectstorage_object/object/objectstorage_object.tf line 6, in resource "oci_objectstorage_object" "objectstorage_object":
[...]

PR #1200, opened 2020-10-09
Avoid chunked HTTP requests for Object Storage

skinowski:
In go net/http, the recommend way to signal an empty body
is using http.NoBody. (or leaving the body nil.) With this
patch, for empty strings and empty body cases, we use
http.NoBody explicitly to avoid any accidental chunked
transfer encoding in HTTP requests.

Before this change, the issue can be identified by the presence
of "Transfer-Encoding: chunked" header and absence of "Content-Length"
header in the requests.

Signed-off-by: Tolga Ceylan <tolga.ceylan@oracle.com>

### DIFF
--- a/oci/objectstorage_object_resource.go
+++ b/oci/objectstorage_object_resource.go
@@ -12,6 +12,7 @@ import (
 	"bytes"
 	"io/ioutil"
 	"log"
+	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


skinowski:
Fixes #1191

skinowski:
Moving this PR to Oracle internal.

PR #1378, opened 2021-05-25
Fix wrong connection strings in lookup function of autonomous_database.tf example file

leonyk:
This is the fix for submitted issue of https://github.com/terraform-providers/terraform-provider-oci/issues/1377


Maxrovr:
This will not fix the issue mentioned

Issue #1258, opened 2020-12-29
Support OCI_CLI_CONFIG_FILE for provider configuratiom


rbair23:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

In my CI/CD configuration I would like to dynamically generate the OCI config file and use OCI_CLI_CONFIG_FILE to refer to it, so that the provider will use that generated config. It appears that the provider only supports ~/.oci/config right now.

sanyer:
Seems like terraform-provider-oci relies on oci-sdk-go mechnism of falling back to filename specified in `OCI_CONFIG_FILE` if `~/.oci/config` wasn't found. But no, it's no possible to use custom config_file path :-(

ravinitp:
We are very sorry that we couldn't respond to each and every issue reported on GitHub. 
Although we have refined the process to prioritize customer issues on GitHub, since this issue was reported a while ago, there is a good chance it may have been fixed in the latest version of Terraform Provider OCI. 
If you are still experiencing this issue, please create a new issue and label it as Bug.

sanyer:
Oh wow, it's been a while :-D
Thanks for answering @ravinitp 

I'll check again if this was fixed.

PR #1482, opened 2021-11-12
Add support for arbitrary config file in provider block

sanyer:
Allow provider support other configuration files than ~/.oci/config

Added configuration argument `config_file_path`
Logic of method `getSdkConfigProvider` was updated to support combinations of `config_file_profile` and `config_file_path`:
- `configFile` and `profile` are initiliased with default values
- in case corresponding configuration argument is provided, default value is being replaced with the provided one
- if neither of configuration arguments provided, `oci_common.DefaultConfigProvider()` is used to populate `configProviders`
- if any is provided, `oci_common.CustomProfileConfigProvider()` with extracted values is used

Implements the idea #1258

### DIFF
--- a/oci/export_compartment_test.go
+++ b/oci/export_compartment_test.go
@@ -1494,20 +1494,20 @@ func TestUnitGetExportConfig(t *testing.T) {
 		t.Skip("This run requires you to set TF_HOME_OVERRIDE")
 	}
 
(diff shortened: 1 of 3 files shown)


jotruon:
@sanyer closing this PR as this is outdated

sanyer:
@jotruon do you know if this was implemented in some other PR?

PR #1521, opened 2022-02-01
Always Free example update: Flexible Load balancer and support for all Regions

junior:
- Always Free Load balancer update to flexible from the 10-micro (Not available for new tenancies)
- Oracle Linux 8 image updated to use latest (Current code has the minimum unpatched 7, that can be deprecated soon)
- Support for all regions supported by the image
- Resource name updated to `oci_load_balancer_load_balancer` from the alias `oci_load_balancer`
- Auto generation of the self-signed certificate (Current code has a hard-coded certificated expired in 2021)

### DIFF
--- a/examples/always_free/main.tf
+++ b/examples/always_free/main.tf
@@ -45,12 +45,30 @@ variable "images" {
 
   default = {
     # See https://docs.us-phoenix-1.oraclecloud.com/images/
-    # Oracle-provided image "Oracle-Linux-7.5-2018.10.16-0"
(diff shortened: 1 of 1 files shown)


spsarumugham (approved):
I tested a minor variation of this through TF cloud and it works well.

jotruon:
@junior please follow internal process for reviewing this PR.
Thanks

junior:
After my PR, someone cloned and applied using this PR: https://github.com/oracle/terraform-provider-oci/pull/1534

PR #1615, opened 2022-06-15
reverted changes to core route table that crashed the provider

ravinitp:


internal/globalvar/version.go
```diff
@@ -7,7 +7,7 @@ import (
 	"log"
 )
 
-const Version = "4.80.0"
+const Version = "4.80.1"
 const ReleaseDate = "2022-06-16"
```
rashik-bhasin, line 11: release date need to change

PR #1763, opened 2023-02-13
docs > containerengine_node_pool > fix

europ:
* Enhanced example (removed deprecated arguments, added example value into variables to remove useless variable assignment, etc.) 
* Additional information added for a few arguments in the argument reference section

### DIFF
--- a/website/docs/r/containerengine_node_pool.html.markdown
+++ b/website/docs/r/containerengine_node_pool.html.markdown
@@ -17,75 +17,146 @@ Create a new node pool.
 ```hcl
 resource "oci_containerengine_node_pool" "test_node_pool" {
 	#Required
-	cluster_id = oci_containerengine_cluster.test_cluster.id
-	compartment_id = var.compartment_id
-	name = var.node_pool_name
-	node_shape = var.node_pool_node_shape
+	cluster_id = "ocid1.cluster.oc1.phx.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+
+	#Required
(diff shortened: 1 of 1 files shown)


website/docs/r/containerengine_node_pool.html.markdown
```diff
@@ -17,75 +17,146 @@ Create a new node pool.
 ```hcl
 resource "oci_containerengine_node_pool" "test_node_pool" {
 	#Required
-	cluster_id = oci_containerengine_cluster.test_cluster.id
-	compartment_id = var.compartment_id
-	name = var.node_pool_name
-	node_shape = var.node_pool_node_shape
+	cluster_id = "ocid1.cluster.oc1.phx.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```
ravinitp, line 20: Please avoid using any sensitive value in code/documentation.

ravinitp: You can use terraform variable for reference.

ravinitp (changes requested):

website/docs/r/containerengine_node_pool.html.markdown
europ: The format visibility is not considered as a sensitive information (see https://docs.oracle.com/en-us/iaas/Content/General/Concepts/identifiers.htm) plus, I have intentionally censored the value by using `a` character only, which is again - not a sensitive information.

If I would use variable as you suggest, the key information is lost because you will not know if it is a string, list, map, etc.

BAD (as it is):
```tf
specific_key = var.specific_key
```
GOOD (as it should be):
```tf
specific_key = "this is a valid value example"
```

ravinitp (commented):
Looks good

Issue #2498, opened 2025-12-22
terraform EXSI Host OCVS OCI : current_commitment and next_commitment variables do not not apply


lingamshravanthi:
<img width="720" height="216" alt="Image" src="https://github.com/user-attachments/assets/c34c0114-ac2e-4d29-8e67-a616a433bcdf" />

We are checking for EXSI Additional Host in terraform, please find the issue below as [current_commitment]=“ONE_YEAR” value shows up error. It is the same error for next_commitment input as well.
╷
│ Error: Value for unconfigurable attribute
│ 
│   with oci_ocvp_esxi_host.aer_ath_ocvs_sddc_cluster_host_4,
│   on aer_ath_ocvs_sddc_cluster_host_4.tf line 8, in resource "oci_ocvp_esxi_host" "aer_ath_ocvs_sddc_cluster_host_4":
│    8:     current_commitment             = "ONE_YEAR"
│ 
│ Can't configure a value for "current_commitment": its value will be decided
│ automatically based on the result of applying this configuration.

tf-oci-pub:
Thank you for reporting the issue. We observed the affected resources are not provided in the description or it's incorrect. We request you to add it in issue description as mentioned in below format.
Example: affected_resources = oci_core_instance , oci_core_instances
 
If it's not related to any particular resource then mention affected resource as terraform.
Example: affected_resources = terraform
 
As this works through automation, request you to follow exact syntax.

tf-oci-pub:
Thank you for reporting the issue. We have raised an internal ticket to track this. Our service engineers will get back to you.

RichardORCL:
I created a fix for this and tested it successfully. Created pull request: https://github.com/oracle/terraform-provider-oci/pull/2552

Need someone to review it.

tf-oci-pub:
Thank you for reporting the issue. We observed the affected resources are not provided in the description or it's incorrect. We request you to add it in issue description as mentioned in below format.
Example: affected_resources = oci_core_instance , oci_core_instances
 
If it's not related to any particular resource then mention affected resource as terraform.
Example: affected_resources = terraform
 
As this works through automation, request you to follow exact syntax.

Richard12215:
Hello, should be fixed with https://github.com/oracle/terraform-provider-oci/releases/tag/v8.17.0

PR #2552, opened 2026-05-05
FIX OCVP: adding ESXi node with commitment field

RichardORCL:
### Fix for OCVP Service - ESXi Host creation not allowing to specify current and next commitments

See issue: terraform EXSI Host OCVS OCI : current_commitment and next_commitment variables do not not apply **#2498**

### Summary

Fixes the Terraform provider schema for oci_ocvp_esxi_host so current_commitment and next_commitment can be set in configuration. They were incorrectly declared as read-only (Computed only), which caused Terraform to fail with “Value for unconfigurable attribute” even though the create/update logic already passed these fields to the OCVP API (matching CLI behavior and the resource documentation).

### Problem

Users configuring current_commitment (e.g. HOUR) hit: 
`Can’t configure a value for "current_commitment": its value will be decided automatically…`

The underlying Create/Update paths already read current_commitment and next_commitment from ResourceData and map them to the SDK request, so the failure was a schema vs. implementation mismatch, not an API limitation.

### Solution

current_commitment: Mark as Optional and Computed, ForceNew (create-time input, value then comes from the API), with diff suppression aligned with current_sku so post-create API normalization does not force unnecessary churn. ConflictsWith current_sku so the deprecated and new arguments are not used together.

next_commitment: Mark as Optional and Computed (updatable input, also reflected from read), and ConflictsWith next_sku for the same reason.
current_sku / next_sku: Extend ConflictsWith to include current_commitment / next_commitment respectively.

### Notes for reviewers

current_commitment intentionally does not mirror current_sku’s conflict with cluster_id, so the documented pattern cluster_id + current_commitment remains valid.

No change to API call shape—only schema alignment with existing CRUD code.

Tested successfully on internal OCVS environment

### DIFF
--- a/internal/service/ocvp/ocvp_esxi_host_resource.go
+++ b/internal/service/ocvp/ocvp_esxi_host_resource.go
@@ -83,8 +83,22 @@ func OcvpEsxiHostResource() *schema.Resource {
 					}
 					return true
 				},
-				ConflictsWith: []string{"cluster_id", "esxi_software_version"},
+				ConflictsWith: []string{"cluster_id", "esxi_software_version", "current_commitment"},
 				Deprecated:    tfresource.FieldDeprecated("current_sku"),
+			},
+			"current_commitment": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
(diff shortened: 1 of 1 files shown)


tf-oci-pub:

PR #2623, opened 2026-09-17
core: coalesce concurrent NSG security rule reads via singleflight (f…

### LINKED ISSUE #2609: oci_core_network_security_group_security_rule: Read lists the NSG's entire rule set once per rule, producing O(N²) API traffic
# oci_core_network_security_group_security_rule: Read lists the NSG's entire rule set once per rule, producing O(N^2) API traffic

### Community Note

* Please vote on this issue by adding a 👍 reaction to the original issue to help us prioritize
* Please do not leave "+1" comments

### Terraform CLI and Provider Versions

* Terraform: v1.15.9
* Provider: oracle/oci v7.32.0

### Affected Resource

* `oci_core_network_security_group_security_rule`

### Summary

`CoreSecurityRuleResourceCrud.Get()` reads a single security rule by calling `ListNetworkSecurityGroupSecurityRules` for the **entire parent NSG**, paging through every rule, then linear-scanning the result for its own ID.

Because Terraform refreshes each rule as an independent resource, an NSG holding *N* rules performs *N* full list traversals per plan  -  transferring **N^2 rule records** in order to read *N* rules. There is no caching or in-flight request coalescing.

This is invisible on small configurations but dominates at scale. On our configuration it accounts for roughly **60% of total plan time**.

### Measured impact

One `terraform plan -refresh=true -parallelism=40` over 34,531 resources:

| Metric | Value |
|---|---|
| Total resources refreshed | 34,531 |
| Of which `oci_core_network_security_group_security_rule` | **20,745 (60.1%)** |

Refresh throughput, sampled from plan-output timestamps in 30s buckets:

| Phase | Throughput |
|---|---|
| Opening burst (VCNs, subnets, route tables, DRG objects) | **92.9 refreshes/s** |
| Sustained 9-minute stretch containing *only* NSG security rules | **4.7 refreshes/s** |

NSG security rule reads are therefore roughly **20x slower per resource** than every other resource type in the same plan, at identical parallelism. This was not rate limiting  -  the run completed with **zero HTTP 429s and zero errors**.

Taking the 19,671 of those rules that we can attribute to a specific parent NSG:

| Metric | Value |
|---|---|
| Distinct NSGs owning them | 1,480 |
| Rules per NSG | mean 13.3, median 6, **max 120** |
| Rule records required (one list per NSG) | 19,671 |
| Rule records actually transferred (sum of N^2 per NSG) | **835,345** |
| **Redundancy factor** | **~42x** |

The cost concentrates in a minority of NSGs, as expected for an O(N^2) effect:

| Rules per NSG | NSGs | Records transferred (sum of N^2) |
|---|---|---|
| 1-5 | 730 | 5,838 |
| 6-20 | 454 | 56,127 |
| 21-50 | 177 | 143,457 |
| 51-100 | 99 | 366,807 |
| 101-200 | 20 | 263,116 |

The 20 NSGs holding more than 100 rules generate **31% of all traffic while holding ~12% of the rules**.

### Root cause

[`internal/service/core/core_network_security_group_security_rule_resource.go:381`](https://github.com/oracle/terraform-provider-oci/blob/master/internal/service/core/core_network_security_group_security_rule_resource.go#L381):

```go
func (s *CoreSecurityRuleResourceCrud) Get() error {
    request := oci_core.ListNetworkSecurityGroupSecurityRulesRequest{}
    // ... resolve NSG id from the composite id ...

    response, err := s.Client.ListNetworkSecurityGroupSecurityRules(context.Background(), request)
    if err != nil { return err }
    var rules []oci_core.SecurityRule
    rules = response.Items
    request.Page = response.OpcNextPage

    for request.Page != nil {   // pages through EVERY rule in the NSG
        listResponse, err := s.Client.ListNetworkSecurityGroupSecurityRules(context.Background(), request)
        if err != nil { return err }
        rules = append(rules, listResponse.Items...)
        request.Page = listResponse.OpcNextPage
    }

    for _, r := range rules {   // then linear-scans for its own ID
        if *r.Id == s.D.Id() { s.Res = &r; break }
    }
    // ...
}
```

We understand why it is written this way: the Core API exposes no `GetNetworkSecurityGroupSecurityRule`, only `Add`/`List`/`Remove`/`Update`, so listing is the only way to read a single rule. The issue is that the list result is discarded after matching one ID, and so is recomputed for every sibling rule in the same NSG.

### Suggested direction (not a patch  -  happy to defer to your preferred approach)

Terraform refreshes an NSG's rules concurrently, so the redundant traversals are largely simultaneous. Request coalescing therefore recovers much of the cost without introducing a cache: `singleflight` collapses only calls already in flight at the same moment, and drops the entry on completion, so there is no cross-operation cache to invalidate.

**Preconditions we checked before suggesting this**, since coalescing is only safe if a write is never followed by a read-back through the shared call:

* `Get()` is **refresh-only** for this resource. `Create()` populates `s.Res` from the `AddNetworkSecurityGroupSecurityRules` response (line 373) and `Update()` from its own response (line 523); neither calls `Get()`.
* `CreateResource`/`UpdateResource` in `internal/tfresource/crud_helpers.go` call `sync.SetData()` after the write, not `sync.Get()`. They call `waitForStateRefresh` only for resources implementing `StatefullyCreatedResource`, which this CRUD does not implement.
* `SetData()` only reads from `s.Res`. That matters because coalesced callers would share one backing slice and `oci_core.SecurityRule` carries pointer fields (`*string`, `*TcpOptions`, ...)  -  a mutation through any of them would corrupt sibling resources. Read-only today, but this is incidental rather than structural, so it likely warrants a comment at the call site.

If any of those assumptions is wrong on a path we missed, please say so  -  that would change the analysis.

**One behavioural change we want to be explicit about:** error fate becomes shared. Today each rule's list call retries independently; with coalescing only the leader's call runs and retries, and followers inherit its outcome, so a single transient failure fails N resources rather than one. Relatedly, only the leader's `RetryPolicy` applies  -  uniform today, since `Get()` hardcodes `GetRetryPolicy(false, "core")`, but it is a coupling that does not exist now. We think this is an acceptable trade for a 42x traffic reduction, but it is a real change and worth a maintainer's judgement.

Implementation note: only the list-and-paginate block can be wrapped. `Get()` mutates per-caller state beforehand (`s.D.SetId(securityRuleId)`, `s.D.Set("network_security_group_id", ...)`), so wrapping the whole function would assign the leader's ID to every coalesced rule.

```go
// package-level
var nsgRulesGroup singleflight.Group   // golang.org/x/sync/singleflight

// in Get(), in place of the List + pagination block:
v, err, _ := nsgRulesGroup.Do(networkSecurityGroupId, func() (interface{}, error) {
    return listAllNsgRules(s.Client, networkSecurityGroupId) // existing pagination loop, extracted
})
if err != nil { return err }
rules := v.([]oci_core.SecurityRule)
```

Because the benefit is bounded by how many reads are concurrent, it scales with `-parallelism`:

| Configuration | List traversals for a 120-rule NSG |
|---|---|
| today | 120 |
| coalescing, `-parallelism=10` | ~12 |
| coalescing, `-parallelism=40` | ~3 |
| coalescing + refresh-scoped cache | 1 |

A short-lived per-NSG cache scoped to the refresh walk (invalidated on any `Add`/`Update`/`Remove` against that NSG) would capture the full ~42x, but it introduces invalidation semantics  -  and unlike coalescing it would be exposed to the write-then-read-back hazard above if `Get()` ever became reachable from the apply path. Coalescing therefore seems the lower-risk starting point. We would be glad to see either.

The same pattern applies to any resource whose `Get()` is implemented as a filtered list over a shared parent collection, so a reusable helper may be worth more than a fix local to this resource.

### Secondary: page size is never set

`request.Limit` is not set in `Get()`, so pagination falls back to the service default. For NSGs with more than 100 rules this multiplies HTTP round trips per traversal on top of the redundancy above. Setting an explicit maximum `Limit` would reduce round trips independently of the caching question.

### Steps to reproduce

1. Create an NSG with ~120 security rules, each as a separate `oci_core_network_security_group_security_rule` resource.
2. `terraform apply`, then `terraform plan -refresh=true`.
3. With `TF_LOG=DEBUG`, observe `ListNetworkSecurityGroupSecurityRules` issued once per rule (~120 times, more with pagination), each response carrying the full 120-rule set.

**Expected:** the NSG's rule list is retrieved once and shared across the rules being refreshed.
**Actual:** retrieved once per rule, transferring N^2 rule records.

Cozm1n1412:
### Description
Fixes #2609.

Currently, `CoreSecurityRuleResourceCrud.Get()` fetches the entire rule set of a parent NSG for every individual security rule during plan/refresh. Under default Terraform parallelism, this causes $O(N^2)$ API traffic and severe throttling/latency on large NSGs.

### Changes
- Integrated `golang.org/x/sync/singleflight` to deduplicate and coalesce concurrent `ListNetworkSecurityGroupSecurityRules` requests targeting the same `network_security_group_id`.
- Set `req.Limit = common.Int(100)` on the listing request to minimize pagination roundtrips.
- Concurrent rule reads now share the aggregated in-flight slice, collapsing network calls to $O(1)$ per NSG refresh window while keeping memory access read-only.


tf-oci-pub:

Cozm1n1412:
Thanks for the update! Looking forward to the local test results. Please let me know if any adjustments or additional test cases are needed once the team reviews the changes.

